PR #1574, opened 2023-01-10
Remove certain filetypes from get_pedigree_file_path

xiamaz:
Adds a blacklist of certain file suffixes, so these are not considered as a valid BAM or VCF when getting a pedigree path.


xiamaz:
@mikkonie This is just a proposed solution, it would be great if we could work this into a workable solution.

mikkonie:
This is a good suggestion. I will extend it a little, add proper testing and deploy into production ASAP. See issue #1575.

PR #1627, opened 2023-04-05
Add CyTOF assay plugin

### LINKED ISSUE #1642: Add a new samplesheet assay plugin for mass cytometry data
Mass cytometry data is quite similar to scRNA-seq data, but differs in the details. For handling these kind of datasets, a dedicated assayapp is needed.

## Requirements
1. Multiple samples are measured as multiplexed pools, so SODAR should point all of these samples to one iRODS collection using the "last material node" logic of the RNA_sequencing plugin.
2. Various protocol parameters are too complex for direct inclusion into the ISA-tab and need to be linked from the MiscFiles collection. The plugin should provide auto-linking of these resources similar to the ms_meta plugin.
3. Various protocols generate report files which should be linked from the respective multiplexed "run" collection.

sellth:
Combination of dna_sequencing & meta_ms assay plugins. 
- creates collections based on AssayName of (last) process named "Mass cytometry"
- creates simple links for antibody panel & barcode key to MiscFiles collection
- creates simple links for report files to respective "AssayName" collections

Fixes #1642.

### DIFF
--- a/samplesheets/assayapps/cytof/plugins.py
+++ b/samplesheets/assayapps/cytof/plugins.py
@@ -19,7 +19,7 @@ class SampleSheetAssayPlugin(SampleSheetAssayPluginPoint):
     name = 'samplesheets_assay_cytof'
 
     #: Title
-    title = 'Mass Cytometry Assay plugin'
+    title = 'Mass Cytometry Assay Plugin'
 
     #: App name for dynamic reference to app in e.g. caching
     app_name = APP_NAME
--- a/samplesheets/assayapps/cytof/plugins.py
+++ b/samplesheets/assayapps/cytof/plugins.py
@@ -6,6 +6,7 @@ from django.conf import settings
 from samplesheets.plugins import SampleSheetAssayPluginPoint
 from samplesheets.utils import get_last_material_name
 from samplesheets.views import MISC_FILES_COLL
+from samplesheets.rendering import SIMPLE_LINK_TEMPLATE
 
 # Local constants
 APP_NAME = 'samplesheets.assayapps.cytof'
@@ -98,8 +99,9 @@ class SampleSheetAssayPlugin(SampleSheetAssayPluginPoint):
                 )
                 and row[i]['value']
             ):
-                row[i]['link'] = (
-                    base_url + '/' + MISC_FILES_COLL + '/' + row[i]['value']
+                row[i]['value'] = SIMPLE_LINK_TEMPLATE.format(
+                    label = row[i]['value'],
+                    url = base_url + '/' + MISC_FILES_COLL + '/' + row[i]['value'],
                 )
 
             # Report file links within processes
@@ -108,8 +110,9 @@ class SampleSheetAssayPlugin(SampleSheetAssayPluginPoint):
                 and header['value'].lower() == 'report file'
                 and row[i]['value']
             ):
-                row[i]['link'] = (
-                    base_url + '/' + last_material_name + '/' + row[i]['value']
+                row[i]['value'] = SIMPLE_LINK_TEMPLATE.format(
+                    label = row[i]['value'],
+                    url = base_url + '/' + last_material_name + '/' + row[i]['value'],
                 )
 
         return row
--- a/samplesheets/assayapps/cytof/plugins.py
+++ b/samplesheets/assayapps/cytof/plugins.py
@@ -1,0 +1,125 @@
+"""Assay app plugin for samplesheets"""
+
+from django.conf import settings
+
+# Projectroles dependency
+from samplesheets.plugins import SampleSheetAssayPluginPoint
+from samplesheets.utils import get_last_material_name
+from samplesheets.views import MISC_FILES_COLL
+
+# Local constants
+APP_NAME = 'samplesheets.assayapps.cytof'
+
+
+class SampleSheetAssayPlugin(SampleSheetAssayPluginPoint):
+    """Plugin for mass cytometry assays in sample sheets"""
+
+    #: Name (used in code and as unique idenfitier)
+    name = 'samplesheets_assay_cytof'
+
+    #: Title
+    title = 'Mass Cytometry Assay plugin'
+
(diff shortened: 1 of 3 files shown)


mikkonie:
Always base pull requests on the `dev` branch, that is the branch currently being worked on.

sellth:
Sorry, my bad. My local branch is based on dev, I just didn't pay attention when creating the pull request.

samplesheets/assayapps/cytof/plugins.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+class SampleSheetAssayPlugin(SampleSheetAssayPluginPoint):
+    """Plugin for mass cytometry assays in sample sheets"""
+
+    #: Name (used in code and as unique idenfitier)
+    name = 'samplesheets_assay_cytof'
+
+    #: Title
+    title = 'Mass Cytometry Assay plugin'
```
mikkonie: It's customary to have these titles in title case and prefixed with `Sample Sheets`, in this case that would be `Sample Sheets Mass Cytometry Assay Plugin`.

sellth: That's in the description field though, right?

sellth:
This is now ready to be merged.

sellth:
After some contemplating about the nature of metadata and the ISA-tab standard, I now consider using PROCESS."Assay Name" as a multiplexing key instead of the last_material_node logic. This would make a process called "Mass cytometry" mandatory for all CyTOF projects though.

related: #1660

sellth:
Rebased to latest dev.

samplesheets/assayapps/cytof/plugins.py
```diff
@@ -0,0 +1,157 @@
+"""Assay app plugin for samplesheets"""
+
+from django.conf import settings
+
+# Projectroles dependency
```
mikkonie: Since all the imports come from samplesheets, this comment can be omitted.

samplesheets/assayapps/cytof/plugins.py
```diff
@@ -0,0 +1,157 @@
...
+from samplesheets.plugins import SampleSheetAssayPluginPoint
+from samplesheets.utils import get_top_header
+from samplesheets.views import MISC_FILES_COLL
+from samplesheets.rendering import SIMPLE_LINK_TEMPLATE
+
+# Local constants
```
mikkonie, line 11: The convention is to have two blank lines between imports and constants.

samplesheets/assayapps/cytof/plugins.py
```diff
@@ -0,0 +1,157 @@
...
+        :param assay: Assay object
+        :param assay_path: Root path for assay
+        :return: String with full iRODS path or None
+        """
+        # Get the value of Mass cytometry Assay Name column
+        mc_assay_name = self.__get_mc_assay_name(row, table)
+        if mc_assay_name:
+            return assay_path + '/' + mc_assay_name
+
+        return None
```
mikkonie: The general convention is to avoid having a single return line hanging at the end of a function, to keep code condensed. You might still see bits of old code in SODAR where this is done, but I'm trying to iron those out :)

samplesheets/assayapps/cytof/plugins.py
```diff
@@ -0,0 +1,157 @@
...
+    def update_row(self, row, table, assay):
+        """
+        Update render table row with e.g. links. Return the modified row.
+
+        :param row: Original row (list of dicts)
+        :param table: Full table (dict)
+        :param assay: Assay object
+        :return: List of dicts
+        """
+
```
mikkonie: We should also avoid blank lines between the docstring and the start of a function.

samplesheets/assayapps/cytof/plugins.py
```diff
@@ -0,0 +1,157 @@
...
+                header['obj_cls'] == 'Process'
+                and (
+                    header['value'].lower() == 'antibody panel'
+                    or header['value'].lower() == 'barcode key'
+                )
+                and row[i]['value']
+            ):
+                row[i]['value'] = SIMPLE_LINK_TEMPLATE.format(
+                    label=row[i]['value'],
+                    url=f"{base_url}/{MISC_FILES_COLL}/{row[i]['value']}",
```
mikkonie: For keeping with SODAR conventions these should be formatted as `'string'.format()`.

samplesheets/assayapps/cytof/plugins.py
```diff
@@ -0,0 +1,157 @@
...
+        :param table: Full table with headers (dict returned by
+                      SampleSheetTableBuilder)
+        """
+        name = None
+        span_end = len(row)
+        for i in range(len(row)):
+            cell = row[i]
+            header = table['field_header'][i]
+            if (
+                header['obj_cls'] == 'Process'
```
mikkonie, line 63: Nice one, I forgot we can access the object class at this point. That makes it simpler than I recalled :)

mikkonie (commented):
This looks good, only had some minor comments on style/conventions.

samplesheets/assayapps/cytof/plugins.py
sellth: Simple string concatenation is used by other assayapps, so I've opted for that now. I've initially used the f_string because it's the fastest option and we iterate through the entire table here.

samplesheets/assayapps/cytof/plugins.py
sellth: Yeah, that one's actually doing nothing, right?

samplesheets/assayapps/cytof/plugins.py
sellth: I've copied from the best. ;-)

sellth:
Comments resolved.

samplesheets/assayapps/cytof/plugins.py
sellth: But I guess, we only modify a few cells…

PR #1629, opened 2023-04-06
 set proper permissions for server.key

sellth:
Right now server.key is owned by the local user or root and has default permissions of 600. From within Docker iRODS wants to access this file via the internal user irods and fails. This is mostly (only?) noticeable when trying to handshake for a Davrods connection, which results in an internal server error message and this in the irods container:

`ERROR: sslInit: couldn't read key file. SSL error: error:0200100D:system library:fopen:Permission denied`

To resolve this (unless there is an easier way) we should instruct the user to set at least 644 permissions on the server.key file.


mikkonie:
Good call, thanks for submitting this pr.

PR #1630, opened 2023-04-14
1619 add rest api endpoint to retrieve project irods file list

### LINKED ISSUE #1619: Add REST API endpoint to retrieve project iRODS file list
This came up in a discussion with Nicolai. To me, it seems like a rather useful feature which is also trivial to implement. As SODAR uses iRODS admin access in the backend, we can do this with a quick query via taskflowbackend.

Will write complete spec later, but putting down the main points now.

## Spec

- Add new API view `ProjectIrodsFileListAPIView` in `samplesheets.views_api`
- Get project UUID as url kwarg as usual
- Required permission for view is `samplesheets.view_sheet`
- Use `irodsbackend.get_objects()` to query for files
    * Query for files under project sample data repository (no landing zones)
        + Get iRODS path to query with `irodsbackend.get_sample_path(project)`
    * Default args for the method (no .md5 files returned, etc)
    * This will return full paths for each file
    * Provide `irods_data` as the resulting data for the API view
- Add tests
    * Skip permission tests for now (I have to do some setup for samplesheets permission tests for iRODS)
    * View tests under `samplesheets.tests.test_views_api_taskflow`
- Add API view docs to `api_samplesheets.rst` using `autoclass`
    * Ensure proper syntax is used in API view docstring, see other API views as an example

gromdimon:
### TODO:
- [x] Add new API view
- [x] Add tests
- [x] Add docs


.gitignore
```diff
@@ -98,6 +98,7 @@ flamegraph.pl
 
 # Celery
 celerybeat-schedule
+celerybeat-schedule.db
```
gromdimon, line 101: My celery creates a new schedule file, so I think it should be added to .gitignore

samplesheets/tests/test_views_api.py
```diff
@@ -17,7 +17,9 @@
 
 # Landingzones dependency
 from landingzones.models import LandingZone
-from landingzones.tests.test_models import LandingZoneMixin
+from landingzones.tests.test_models import (
+    LandingZoneMixin,
```
mikkonie: Removing the `,` should make this fit back to a single line again :)

samplesheets/views_api.py
```diff
@@ -421,3 +422,49 @@ def get(self, request, **kwargs):
...
+    permission_required = 'samplesheets.view_sheet'
+
+    def get(self, request, *args, **kwargs):
+        if not settings.ENABLE_IRODS:
+            raise APIException('iRODS not enabled')
+        irods_backend = get_backend_api('omics_irods')
+        if not irods_backend:
+            raise APIException('iRODS backend not enabled')
+        project = self.get_project(request=request, kwargs=kwargs)
+        self.check_object_permissions(request, project)
```
mikkonie: This should not be needed as the class is based on `SODARAPIBaseProjectMixin` and  `permission_required` has been set.

samplesheets/views_api.py
```diff
@@ -421,3 +422,49 @@ def get(self, request, **kwargs):
             except Exception as ex:
                 return Response(str(ex), status=500)
         return Response(ret, status=200)
+
+
+class ProjectIrodsFileListAPIView(
+    SODARAPIBaseProjectMixin, SODARAPIBaseMixin, APIView
```
mikkonie: `SODARAPIBaseProjectMixin` itself is based on `SODARAPIBaseMixin`, so the latter dependency is redundant.

samplesheets/views_api.py
```diff
@@ -421,3 +422,49 @@ def get(self, request, **kwargs):
...
+    http_method_names = ['get']
+    permission_required = 'samplesheets.view_sheet'
+
+    def get(self, request, *args, **kwargs):
+        if not settings.ENABLE_IRODS:
+            raise APIException('iRODS not enabled')
+        irods_backend = get_backend_api('omics_irods')
+        if not irods_backend:
+            raise APIException('iRODS backend not enabled')
+        project = self.get_project(request=request, kwargs=kwargs)
```
mikkonie: Just `self.get_project()` should be enough.

samplesheets/views_api.py
```diff
@@ -421,3 +422,49 @@ def get(self, request, **kwargs):
...
+        self.check_object_permissions(request, project)
+        path = request.query_params.get('path')
```
mikkonie: There was no requirement in the spec for setting a custom path. This can be removed as, at least for now, we are interested in the full file listing under the sample repository of the project.

Also, the method here is not protected against iRODS paths containing `/../`. That could be used to get access to read a collection outside of the bounds of the project where the user might not have actual rights. They could not actually access the files, just get a file listing, but still it is a serious security issue. While thinking of this I realized that the same thing actually works for a couple of existing views too. I'll add a separate ticket for sanitizing path input.

samplesheets/views_api.py
```diff
@@ -421,3 +422,49 @@ def get(self, request, **kwargs):
...
+        if not path:
+            path = irods_backend.get_sample_path(project)
+        else:
+            path = os.path.join(irods_backend.get_sample_path(project), path)
+        try:
+            with irods_backend.get_session() as irods:
+                irods_data = irods_backend.get_objects(irods, path)
+        except Exception as ex:
+            return Response(
+                {'detail': 'Unable to connect to iRODS: {}'.format(ex)},
```
mikkonie: We actually can't know that any exception returned by `get_objects()` would mean a connection to iRODS can not be made. There may be other things going wrong. Instead, report this back as e.g. "Exception querying iRODS objects".

samplesheets/tests/test_views_api.py
```diff
@@ -17,7 +17,9 @@
```
mikkonie: Tests for this view must go under `samplesheets.tests.test_views_api_taskflow`, as written in the spec.

samplesheets/tests/test_views_api.py
```diff
@@ -684,3 +693,58 @@ def test_get_isatab(self):
...
+
+    def test_get_no_collection(self):
+        """Test GET request in ProjectIrodsFileListAPIView without collection"""
+        url = reverse(
+            'samplesheets:api_irods_files',
+            kwargs={'project': self.project.sodar_uuid},
+        )
+        with self.login(self.user):
+            response = self.client.get(url)
+        self.assertEqual(response.status_code, 500)
```
mikkonie: I believe in this case we should return `404` instead of `500`.

samplesheets/tests/test_views_api.py
```diff
@@ -684,3 +693,58 @@ def test_get_isatab(self):
...
+        self.assertEqual(
+            response.data['detail'],
+            'Unable to connect to iRODS: iRODS collection not found',
```
mikkonie: It would be best to set up these strings as constants to make testing less prone to errors when we change things.

samplesheets/tests/test_views_api.py
```diff
@@ -684,3 +693,58 @@ def test_get_isatab(self):
...
+        )
+
+    def test_get(self):
```
mikkonie: Also add test with files uploaded. You should find examples of tests with uploaded files in `test_views_api_taskflow`, in which these tests should also reside.

samplesheets/urls.py
```diff
@@ -193,6 +193,11 @@
         view=samplesheets.views_api.RemoteSheetGetAPIView.as_view(),
         name='api_remote_get',
     ),
+    url(
+        regex=r'^api/irods/files/(?P<project>[0-9a-f-]+)',
```
mikkonie: To make this more consistent with API view URL paths, we should probably use `api/file/list/` etc.

samplesheets/views_api.py
```diff
@@ -421,3 +421,40 @@ def get(self, request, **kwargs):
             except Exception as ex:
                 return Response(str(ex), status=500)
         return Response(ret, status=200)
+
+
+class ProjectIrodsFileListAPIView(SODARAPIBaseProjectMixin, APIView):
+    """
+    Return a list of files in the project's sample data repository.
+
+    **URL:** ``/samplesheets/api/irods/files/{Project.sodar_uuid}``
```
mikkonie: New URL should be updated here too, of course.

samplesheets/urls.py
```diff
@@ -193,6 +193,11 @@
...
+        view=samplesheets.views_api.ProjectIrodsFileListAPIView.as_view(),
+        name='api_irods_files',
```
mikkonie: Similarly, the URL name should be `api_file_list`.

samplesheets/views_api.py
```diff
@@ -421,3 +421,40 @@ def get(self, request, **kwargs):
...
+            raise APIException('iRODS not enabled')
+        irods_backend = get_backend_api('omics_irods')
+        project = self.get_project()
+        path = irods_backend.get_sample_path(project)
+        try:
+            with irods_backend.get_session() as irods:
+                irods_data = irods_backend.get_objects(irods, path)
+        except Exception as ex:
+            return Response(
+                {'detail': 'Exception queried iRODS objects: {}'.format(ex)},
```
mikkonie: We should already be using a constant instead of a string for the error prefix here, and use that in tests. Also as a bit of proofreading, it could be improved into "Exception querying iRODS objects".

samplesheets/tests/test_views_api_taskflow.py
```diff
@@ -149,7 +156,7 @@ def test_get(self):
         url = reverse('samplesheets:api_file_exists')
         response = self.request_knox(url, data={'checksum': IRODS_FILE_MD5})
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(json.loads(response.content)['status'], False)
+        self.assertEqual(json.loads(response.content)['status'], True)
```
mikkonie: I'm not sure why this one line in an existing test is updated?

samplesheets/tests/test_views_api_taskflow.py
```diff
@@ -185,3 +192,80 @@ def test_get_invalid_checksum(self):
         url = reverse('samplesheets:api_file_exists')
         response = self.request_knox(url, data={'checksum': 'Notvalid MD5!'})
         self.assertEqual(response.status_code, 400)
+
+
+class TestProjectIrodsFileListAPIView(
+    TestSampleSheetAPITaskflowBase,
```
mikkonie: This should fit on one line without the comma.

samplesheets/views_api.py
```diff
@@ -421,3 +422,40 @@ def get(self, request, **kwargs):
...
+    """
+    Return a list of files in the project's sample data repository.
+
+    **URL:** ``/samplesheets/api/file/list/{Project.sodar_uuid}``
+
+    **Methods:** ``GET``
+
+    **Parameters:**
+
+    - ``project``: Project SODAR UUID
```
mikkonie: We can actually remove the "parameters" part from the docstring, as the project UUID is already provided as a part of the URL itself.

PR #1644, opened 2023-04-24
Landingzone update view

gromdimon:
### TODO:
- [x] Add `ZoneUpdateView` to update landing zone
- [x] Add view and permission tests
- [x] Add note about the ability to update landing zones after creation (but only the description) in the manual

### DIFF
--- a/landingzones/templates/landingzones/_zone_buttons.html
+++ b/landingzones/templates/landingzones/_zone_buttons.html
@@ -56,6 +56,14 @@
       Copy Zone UUID
     </a>
 
+    {% if zone.status != 'MOVED' and can_update_all or zone.user == request.user and can_update_own %}
+      <a class="dropdown-item sodar-lz-zone-btn-update"
+         href="{% url 'landingzones:update' landingzone=zone.sodar_uuid %}">
+        <i class="iconify" data-icon="mdi:update"></i>
+        Update Zone
+      </a>
+    {% endif %}
+
     {% if zone.status != 'MOVED' and can_delete_all or zone.user == request.user and can_delete_own %}
       <a class="dropdown-item text-danger sodar-lz-zone-btn-delete"
          href="{% url 'landingzones:delete' landingzone=zone.sodar_uuid %}">
--- a/landingzones/templates/landingzones/_zone_buttons.html
+++ b/landingzones/templates/landingzones/_zone_buttons.html
@@ -59,7 +59,7 @@
     {% if zone.status != 'MOVED' and can_update_all or zone.user == request.user and can_update_own %}
       <a class="dropdown-item sodar-lz-zone-btn-update"
          href="{% url 'landingzones:update' landingzone=zone.sodar_uuid %}">
-        <i class="iconify" data-icon="mdi:update"></i>
+        <i class="iconify" data-icon="mdi:lead-pencil"></i>
         Update Zone
       </a>
     {% endif %}
--- a/landingzones/tests/test_permissions.py
+++ b/landingzones/tests/test_permissions.py
@@ -173,13 +173,16 @@ class TestLandingZonePermissions(TestLandingZonePermissionsBase):
             self.user_delegate,
             self.user_contributor,
         ]
-        # bad_users = [
-        #     self.user_guest,
-        #     self.anonymous,
-        #     self.user_no_roles,
-        # ]
+        bad_users = [
+            self.user_guest,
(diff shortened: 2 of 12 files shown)


gromdimon:
## TODO:
- [x] Check taskflow
- [x] Check permissions

gromdimon:
### After all I have 2 Q:
- [x] Currently the `ZoneUpdateView` doesn't use the taskflow for updates. Should we use it or let it be as is?
- [ ] In the permissions tests I found a difficulty asserting responses from bad users. The thing is, that for anonymous users there is no access even for a redirect url, so theay are being redirectred to home. But with `assert_response()` command I can not check multiple redirect_urls. Should I change this tests as well?

landingzones/templates/landingzones/_zone_buttons.html
```diff
@@ -56,6 +56,14 @@
       Copy Zone UUID
     </a>
 
+    {% if zone.status != 'MOVED' and can_update_all or zone.user == request.user and can_update_own %}
+      <a class="dropdown-item sodar-lz-zone-btn-update"
+         href="{% url 'landingzones:update' landingzone=zone.sodar_uuid %}">
+        <i class="iconify" data-icon="mdi:update"></i>
```
mikkonie: The convention is to use `mdi:lead-pencil` as the icon. I know, I should create that style guide some day.. :)

landingzones/tests/test_permissions.py
```diff
@@ -160,6 +160,80 @@ def test_zone_create_disable(self):
...
+        bad_users = [
+            self.user_guest,
+            self.anonymous,
+            self.user_no_roles,
+        ]
+        self.assert_response(url, good_users, 200)
+        try:
+            self.assert_response(url, bad_users, 302)
+        except AssertionError:
+            pass  # In some cases, the redirect is not properly followed
```
mikkonie: These we should handle. In which cases? What happens?

landingzones/views.py
```diff
@@ -221,6 +222,70 @@ def submit_create(
...
+                'description': zone.description,
+            }
+            tl_event = timeline.add_event(
+                project=zone.project,
+                app_name=APP_NAME,
+                user=user,
+                event_name='zone_update',
+                description='update landing zone {{{}}} for {{{}}} in '
+                '{{{}}}'.format('zone', 'user', 'assay'),
+                status_type='SUBMIT',
```
mikkonie: In this case we are simply updating the LandingZone object synchronously and no taskflow is used. Hence, the status should be set to `OK` here. Right now these appear as stuck in the `SUBMIT` status in timeline:

2023-04_zone_update_tl_bug

landingzones/views.py
```diff
@@ -221,6 +222,70 @@ def submit_create(
             raise ex
 
 
+class ZoneUpdateMixin(ZoneConfigPluginMixin):
+    """Mixin to be used in zone update in UI and REST API views"""
+
+    def submit_update(self, zone, request=None):
```
mikkonie: This can be renamed to e.g. `update_zone()`, as we are not doing any async submission here.

mikkonie (commented):
The spec was written a long time ago and I just noticed there are two major things missing:

- Also make the `user_message` field editable
- Add a REST API view for updating via the API

I've now updated the original spec to reflect this. Sorry for the omissions.

landingzones/tests/test_permissions.py
gromdimon: I pointed this out in the above message because I didn't know how to handle it. The difficulty lies in the fact that, e.g. an anonymous user does not have access to the project that contains the landingzone. The result is that the user is redirected from the url of the target zone to the URL of the project, but since he also does not have permissions to this, he is redirected to the home page. The `assert_response()` method cannot check multiple redirection URLs, although it throws an error and the test fails.

I see the following solutions for that:
- Modify `assert_response()` method to check multiple redirect urls
- Make loops in tests for individual user, login with it, and check redirection

What do you think on that?

mikkonie: I think the best way would be to:

1. Open a ticket in SODAR Core for updating `assert_response()` to handle a redirect chain
2. Implement the `assert_response()` update for v0.13.0
3. Add a ticket in SODAR for updating these tests once we have upgraded to v0.13.0
4. Add a comment in the current test implementation(s) to point to the aforementioned ticket

PR #1656, opened 2023-05-08
Landingzone Update views

### LINKED ISSUE #1267: Enable landing zone updating
Forms code exists for this but for whatever reason, this has been disabled in the UI since the prototype. Couldn't find a reason for it.

The fields which are viable for updating right now are the description and user message.

## Spec

- Add `ZoneUpdateView` to update landing zone
    * Use `LandingZoneForm`
    * Requires permission `landingzones.update_zone_own` or `landingzones.update_zone_all` depending on whether request user owns the zone
- Add links for updating landing zones in zone dropdowns
    * View according to permission as described above: If it's the user's own zone, check for `update_zone_own`, else `update_zone_all`
- Ensure only the `description` and `user_message` fields can be updated in the form
- Add `ZoneUpdateAPIView` in the REST API
    * Same to the UI, only allow updating of `description` and `user_message`
    * Use the serializer, but also call the update method in `ZoneUpdateMixin` to ensure relevant timeline events are created.
- Add view and permission tests
    * These go under `test_views` and `test_permissions`
    * API view permission/view tests under the respective modules, of course
    * No need to have taskflow/iRODS stuff, since we can simply test with a `LandingZone` object in the database
- Add note about the ability to update landing zones after creation (but only the description) in the manual under `app_landingzones_create`

gromdimon:
### TODO:
- [x] Add `ZoneUpdateView` to update landing zone
- [x] Add view and permission tests
- [x] Add note about the ability to update landing zones after creation (but only the description) in the manual

### DIFF
--- a/docs_manual/source/app_landingzones_create.rst
+++ b/docs_manual/source/app_landingzones_create.rst
@@ -71,3 +71,13 @@ see the zone status and move further with file uploads.
 
 The next sections will provide instructions on browsing your landing zones and
 how to proceed with your file uploads.
+
+
+Landing Zone Update
+^^^^^^^^^^^^^^^^^^^
+
+Landing zones can be updated by clicking the :guilabel:`Update Zone` button on the
+dropdown menu on the right hand side of the zone list. This opens up the same
+form as in the creation process, with the fields pre-filled with the current
+values. Currently the only field that can be updated is the description. Once
+you have made your changes, click :guilabel:`Update` to save the changes.
--- a/docs_manual/source/app_landingzones_create.rst
+++ b/docs_manual/source/app_landingzones_create.rst
@@ -79,5 +79,5 @@ Landing Zone Update
 Landing zones can be updated by clicking the :guilabel:`Update Zone` button on the
 dropdown menu on the right hand side of the zone list. This opens up the same
 form as in the creation process, with the fields pre-filled with the current
-values. Currently the only field that can be updated is the description. Once
-you have made your changes, click :guilabel:`Update` to save the changes.
+values. Currently the only fields that can be updated are description and user message.
+Once you have made your changes, click :guilabel:`Update` to save them.
--- a/landingzones/templates/landingzones/landingzone_form.html
+++ b/landingzones/templates/landingzones/landingzone_form.html
@@ -27,7 +27,7 @@
           <i class="iconify" data-icon="mdi:arrow-left-circle"></i> Cancel
         </a>
         <button type="submit" class="btn btn-primary">
-          {% if action == 'create' %}
+          {% if not object.pk %}
             <i class="iconify" data-icon="mdi:plus-thick"></i> Create
           {% else %}
             <i class="iconify" data-icon="mdi:check-bold"></i> Update
--- a/landingzones/templates/landingzones/landingzone_form.html
+++ b/landingzones/templates/landingzones/landingzone_form.html
@@ -3,13 +3,13 @@
 {% load crispy_forms_filters %}
 
 {% block title %}
-  Create Landing Zone in {{ project.title }}
+  {% if action == 'create' %}Create{% else %}Update{% endif %} Landing ababab Zone in {{ project.title }}
 {% endblock title %}
 
 {% block projectroles_extend %}
 
 <div class="container-fluid sodar-subtitle-container">
-  <h3>Create Landing Zone</h3>
+  <h3>{% if action == 'create' %}Create{% else %}Update{% endif %} Landing Zone</h3>
 </div>
 
 <div class="container-fluid sodar-page-container">
@@ -23,7 +23,11 @@
           <i class="iconify" data-icon="mdi:arrow-left-circle"></i> Cancel
         </a>
         <button type="submit" class="btn btn-primary">
-          <i class="iconify" data-icon="mdi:plus-thick"></i> Create
+          {% if action == 'create' %}
+            <i class="iconify" data-icon="mdi:plus-thick"></i> Create
+          {% else %}
+            <i class="iconify" data-icon="mdi:check-bold"></i> Update
+          {% endif %}
         </button>
       </div>
     </div>
--- a/landingzones/templates/landingzones/landingzone_form.html
+++ b/landingzones/templates/landingzones/landingzone_form.html
@@ -3,7 +3,7 @@
 {% load crispy_forms_filters %}
 
 {% block title %}
-  {% if action == 'create' %}Create Landing Zone in {{ project.title }}
+  {% if not object.pk %}Create Landing Zone in {{ project.title }}
   {% else %}Update Landing Zone {% if request.user.username == zone.user.username %}{{ zone.title }}{% else %}{{ zone.user.username }} / {{ zone.title }}{% endif %}
   {% endif %}
 {% endblock title %}
@@ -11,7 +11,7 @@
 {% block projectroles_extend %}
 
 <div class="container-fluid sodar-subtitle-container">
-  <h3>{% if action == 'create' %}Create Landing Zone in {{ project.title }}
+  <h3>{% if not object.pk %}Create Landing Zone in {{ project.title }}
   {% else %}Update Landing Zone {% if request.user.username == zone.user.username %}{{ zone.title }}{% else %}{{ zone.user.username }} / {{ zone.title }}{% endif %}
   {% endif %}</h3>
 </div>
--- a/landingzones/templates/landingzones/landingzone_form.html
+++ b/landingzones/templates/landingzones/landingzone_form.html
@@ -3,13 +3,17 @@
 {% load crispy_forms_filters %}
 
 {% block title %}
-  {% if action == 'create' %}Create{% else %}Update{% endif %} Landing ababab Zone in {{ project.title }}
+  {% if action == 'create' %}Create Landing Zone in {{ project.title }}
+  {% else %}Update Landing Zone {% if request.user.username == zone.user.username %}{{ zone.title }}{% else %}{{ zone.user.username }} / {{ zone.title }}{% endif %}
+  {% endif %}
 {% endblock title %}
 
 {% block projectroles_extend %}
 
 <div class="container-fluid sodar-subtitle-container">
-  <h3>{% if action == 'create' %}Create{% else %}Update{% endif %} Landing Zone</h3>
+  <h3>{% if action == 'create' %}Create Landing Zone in {{ project.title }}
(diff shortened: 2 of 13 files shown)


gromdimon:
Previous PR: #1644

docs_manual/source/app_landingzones_create.rst
```diff
@@ -71,3 +71,13 @@ see the zone status and move further with file uploads.
...
 how to proceed with your file uploads.
+
+
+Landing Zone Update
+^^^^^^^^^^^^^^^^^^^
+
+Landing zones can be updated by clicking the :guilabel:`Update Zone` button on the
+dropdown menu on the right hand side of the zone list. This opens up the same
+form as in the creation process, with the fields pre-filled with the current
+values. Currently the only field that can be updated is the description. Once
```
mikkonie: User message can also be updated now :)

landingzones/templates/landingzones/landingzone_form.html
```diff
@@ -3,13 +3,13 @@
 {% load crispy_forms_filters %}
 
 {% block title %}
-  Create Landing Zone in {{ project.title }}
+  {% if action == 'create' %}Create{% else %}Update{% endif %} Landing ababab Zone in {{ project.title }}
```
mikkonie: If updating, instead of `in {{ project.title }}` we should have the zone title in here. If the zone user is different from the user performing the request, also include the user name in form of `{{ zone.user.username }} / {{ zone.title }}.

landingzones/templates/landingzones/landingzone_form.html
```diff
@@ -3,13 +3,13 @@
...
 {% endblock title %}
 
 {% block projectroles_extend %}
 
 <div class="container-fluid sodar-subtitle-container">
-  <h3>Create Landing Zone</h3>
+  <h3>{% if action == 'create' %}Create{% else %}Update{% endif %} Landing Zone</h3>
```
mikkonie: Same for here as for the page title. Since we are not displaying all the fields in the update form, it's not clear by first glance which zone we are in fact updating. Hence here we should also display the landing zone title (and the user name if zone user is different from request user).

landingzones/tests/test_permissions.py
```diff
@@ -160,6 +160,83 @@ def test_zone_create_disable(self):
         self.assert_response(url, good_users, 200)
         self.assert_response(url, bad_users, 302)
 
+    def test_zone_update(self):
+        """Test ZoneUpdateView default permissions"""
```
mikkonie: We can just have "Test ZoneUpdateView permissions" here.

landingzones/tests/test_permissions_api.py
```diff
@@ -29,7 +31,10 @@
 
 
 class TestLandingZonePermissions(
-    LandingZoneMixin, SampleSheetIOMixin, TestProjectPermissionBase
+    LandingZoneMixin,
+    SampleSheetIOMixin,
+    TestProjectPermissionBase,
+    SODARAPIPermissionTestMixin,
```
mikkonie: Remember, the base class should always come after mixins.

landingzones/tests/test_permissions_api.py
```diff
@@ -127,3 +132,101 @@ def test_retrieve_archive(self):
...
+            self.user_no_roles,
+        ]
+        self.assert_response_api(
+            url,
+            good_users,
+            200,
+            method='PATCH',
+            data=self._get_post_data(),
+            knox=True,
+            req_kwargs={'CONTENT_TYPE': 'application/json'},
```
mikkonie: These content type redefinitions should not be needed. I'm guessing there is something wrong with either the view or test class setup. I'll take a closer look and get back to you on this one.

landingzones/tests/test_views.py
```diff
@@ -173,6 +173,40 @@ def test_render(self):
...
+                    'landingzones:update',
+                    kwargs={'landingzone': self.landing_zone.sodar_uuid},
+                )
+            )
+        self.assertEqual(response.status_code, 200)
+        # Assert form
+        form = response.context['form']
+        self.assertIsNotNone(form)
+        self.assertIsNotNone(form.fields['assay'])
+        self.assertIsNotNone(form.fields['description'])
```
mikkonie, line 195: Make sure to also assert the expected fields are hidden with the `HiddenInput` widget.

landingzones/tests/test_views.py
```diff
@@ -173,6 +173,40 @@ def test_render(self):
...
+        with self.login(self.user):
+            response = self.client.get(
+                reverse(
+                    'landingzones:update',
+                    kwargs={'landingzone': self.landing_zone.sodar_uuid},
+                )
+            )
+        self.assertEqual(response.status_code, 302)
+
+
```
mikkonie, line 291: Make sure to also test a `POST` request for a successful update.

landingzones/tests/test_views_api.py
```diff
@@ -221,3 +221,37 @@ def test_get_locked(self):
...
+
+    def test_patch_title(self):
+        """Test updating title with patch() (should fail)"""
+        url = reverse(
+            'landingzones:api_update',
+            kwargs={'landingzone': self.landing_zone.sodar_uuid},
+        )
+        data = {'assay': str(self.assay.sodar_uuid), 'title': 'New title'}
+        response = self.request_knox(url, method='PATCH', data=data)
+        self.assertEqual(response.status_code, 500)
```
mikkonie: This should return 400 (bad request), not 500 (internal server error). Same with other fields which are not allowed to be edited.

landingzones/views.py
```diff
@@ -221,6 +222,49 @@ def submit_create(
             raise ex
 
 
+class ZoneUpdateMixin(ZoneConfigPluginMixin):
```
mikkonie: To keep with SODAR and SODAR Core tradition, I'd like you to combine this with `ZoneCreateMixin` into `ZoneModifyMixin`. It's logical to have a single helper mixin for both create and update actions. They can then potentially share the same helper methods etc.

landingzones/views.py
```diff
@@ -413,6 +457,12 @@ class ZoneCreateView(
     form_class = LandingZoneForm
     permission_required = 'landingzones.create_zone'
 
+    def get_context_data(self, *args, **kwargs):
+        """Add action to context"""
+        context = super().get_context_data(*args, **kwargs)
+        context['action'] = 'create'
```
mikkonie: This is actually not needed: thanks to using standard create and update views, you can just do `{% if object.pk %}` to check whether we are updating an existing object or creating a new one.

landingzones/views.py
```diff
@@ -480,6 +530,102 @@ def form_valid(self, form):
...
+    form_class = LandingZoneForm
+    slug_url_kwarg = 'landingzone'
+    slug_field = 'sodar_uuid'
+
+    def get_context_data(self, *args, **kwargs):
+        context = super().get_context_data(*args, **kwargs)
+        context['zone'] = LandingZone.objects.get(
+            sodar_uuid=self.kwargs['landingzone']
+        )
+        context['action'] = 'update'
```
mikkonie: Same as in the create view, this is not needed. Just check for `{% if object.pk %}`.

landingzones/views_api.py
```diff
@@ -227,6 +234,62 @@ def perform_create(self, serializer):
             raise APIException('{}{}'.format(ex_msg, ex))
 
 
+class ZoneUpdateAPIView(
+    ZoneUpdateMixin, SODARAPIGenericProjectMixin, UpdateAPIView
+):
+    """
+    Update a landing zone.
```
mikkonie: You could add here that this can be used to update the zone description and user mesasge.

landingzones/views_api.py
```diff
@@ -227,6 +234,62 @@ def perform_create(self, serializer):
...
+    **Methods:** ``PATCH``
+
+    **Parameters:**
+
+    - ``assay``: Assay UUID (string)
+    - ``config_data``: Data for special configuration (JSON, optional)
+    - ``configuration``: Special configuration (string, optional)
+    - ``description``: Landing zone description (string, optional)
+    - ``user_message``: Message displayed to users on successful moving of zone (string, optional)
+    - ``title``: Suffix for the zone title (string, optional)
```
mikkonie: `description` and `user_message` should be the only accepted parameters, trying to update other fields should result in HTTP 400.

landingzones/views_api.py
```diff
@@ -227,6 +234,62 @@ def perform_create(self, serializer):
...
+
+    def perform_update(self, serializer):
+        """
+        Override perform_update() to add timeline event and initiate taskflow.
+        """
+        ex_msg = 'Updating landing zone failed: '
+
+        # Check that only allowed fields are updated
+        if not self._validate_update_fields(serializer):
+            raise APIException('{}Invalid update fields'.format(ex_msg))
```
mikkonie: This should return `ValidationError` instead for that HTTP 400.

landingzones/views_api.py
```diff
@@ -227,6 +234,62 @@ def perform_create(self, serializer):
...
 
+class ZoneUpdateAPIView(
+    ZoneUpdateMixin, SODARAPIGenericProjectMixin, UpdateAPIView
+):
+    """
+    Update a landing zone.
+
+    **URL:** ``/landingzones/api/update/{LandingZone.sodar_uuid}``
+
+    **Methods:** ``PATCH``
```
mikkonie: For consistency we should also support `PUT`, which requires both the `description` and `user_message` fields to be present.

landingzones/tests/test_views_api.py
```diff
@@ -221,3 +221,37 @@ def test_get_locked(self):
         response = self.request_knox(url)
         self.assertEqual(response.status_code, 200)
         self.assertEqual(json.loads(response.content)['status_locked'], True)
+
+
+class TestLandingZoneUpdateAPIView(TestLandingZoneAPIViewsBase):
```
mikkonie, line 226: Also add tests for `PUT` requests once supported.

landingzones/tests/test_permissions_api.py
mikkonie: I fixed this with a new commit to your work branch. The tests were based on `TestProjectPermissionBase` instead of `TestProjectAPIPermissionBase`.

gromdimon: Oh way. That was such a miss, I spent almost 2 days on trying to understand the failure, and finally it was just in Base test class :(

PR #1675, opened 2023-05-12
fix: input widget not accepting multiple files

### LINKED ISSUE #1670: Sheet upload widget only allows single file
Prior to this update it was possible to select multiple files in the upload file dialogue of the "Replace Current Sample Sheets with ISA-Tab" window. Currently only one file can be selected.

This seems due to changes in bd3fec4a52d24b5d716c2207a1a4183853b02604. The new argument *allow_multiple_selected* is not being properly handled by Django i. e. replaced with the HTML "multiple" attribute.

grafik

Replacing *allow_multiple_selected* with *multiple=""* in the browser rescues this phenotype.

### System Information
 - Ubuntu 22.04
 - Firefox 113

sellth:
As per [Django documentation](https://docs.djangoproject.com/en/4.2/topics/http/file-uploads/#uploading-multiple-files) a subclass needs to created to set attribute `allow_multiple_selected=True`. args should remain as `'multiple': True`.

fixes #1670

### DIFF
--- a/samplesheets/forms.py
+++ b/samplesheets/forms.py
@@ -40,12 +40,13 @@ class SheetImportForm(forms.Form):
     directory.
     """
 
+    class MultipleFileInput(forms.ClearableFileInput):
+        allow_multiple_selected = True
+
     file_upload = forms.FileField(
         allow_empty_file=False,
         help_text='Zip archive or ISA-Tab files for a single investigation',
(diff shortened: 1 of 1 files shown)


mikkonie:
Fixed here: https://github.com/bihealth/sodar-server/commit/cc6c74d69a45eb1553c8724a859acf93ac066780

PR #1680, opened 2023-05-16
Add setting to disable study table cache in development

gromdimon:
### TODO:
- [x] Add SHEETS_ENABLE_STUDY_TABLE_CACHE Django setting
- [x] Use setting in get_study_tables()
- [x] Add setting to SHEETS_INFO_SETTINGS
- [x] Assert setting effects cache item creation in TestTableBuilder
- [x] Document setting in admin_settings.rst

### Previous PR: #1668


config/settings/production.py
```diff
@@ -144,6 +144,9 @@
 # ------------------------------------------------------------------------------
 
 
+# SampleSheets
```
mikkonie: Very minor nitpick, but usually this is in form of `# Samplesheets app settings` :)

samplesheets/rendering.py
```diff
@@ -854,17 +856,23 @@ def get_study_tables(self, study):
...
+                    name=item_name,
+                    project=project,
+                )
+                if item and item.data:
+                    logger.debug('Returning cached study tables')
+                    return item.data
+                logger.debug('Cache item "{}" not set'.format(item_name))
+        else:
+            logger.debug(
+                'Study table cache disabled in settings. See STUDY_TABLE_CACHE_ITEM in settings.py. Returning new tables.'
```
mikkonie: Line length over 80 characters :) Sadly black doesn't fix lines with strings, has to be done manually..

samplesheets/tests/test_rendering.py
```diff
@@ -199,6 +199,25 @@ def test_get_study_tables_cache(self):
         study_tables = self.tb.get_study_tables(self.study)
         self.assertEqual(study_tables, cache_item.data)
 
+    def test_get_study_tables_config(self):
```
mikkonie: I'm not sure if I understand the relevance of this test case, as the sheet config isn't really related to the caching.

The test case should be something like this:

- Override cache setting to false
- Manually build study tables in cache with `build_study_tables()`
- Assert the value of a certain study/assay field is X
    * Either in the built tables in cache or the database model, shouldn't matter at this point
- Change the field value to Y in the database model
- Call `get_study_tables()`
    * As the cache setting is false, the cached version should not be returned
- Assert that the field value in the returned tables is Y 

What field to use? I'll leave that up to you :) This is also a good chance to get to know the samplesheets apps models and how they are transformed into the render tables. Ask me if you get stuck on this!

samplesheets/tests/test_rendering.py
```diff
@@ -202,6 +203,35 @@ def test_get_study_tables_cache(self):
         study_tables = self.tb.get_study_tables(self.study)
         self.assertEqual(study_tables, cache_item.data)
 
+    @override_settings(SHEETS_ENABLE_STUDY_TABLE_CACHE=False)
+    def test_get_study_tables_config(self):
+        """Test get_study_tables() with SHEETS_ENABLE_STUDY_TABLE_CACHE=False"""
+        tables = self.tb.get_study_tables(self.study)
+        t_field = tables['study']['top_header'][0]
+        self.assertEqual(t_field['value'], 'Source')
```
mikkonie: We don't need to assert the first top header being called `Source`, that is a given in all of our ISA-Tabs.

samplesheets/tests/test_rendering.py
```diff
@@ -202,6 +203,35 @@ def test_get_study_tables_cache(self):
...
+            0
+        ]
+        self.assertEqual(c_field['header'], 'Source')
+        # Change name in a model
+        characteristics = (
+            GenericMaterial.objects.filter(study=self.study, item_type='SOURCE')
+            .first()
+            .characteristics
+        )
+        characteristics['age']['value'] = '70'
```
mikkonie, line 221: We do need to first assert the original, unmodified value, before we change it.

samplesheets/tests/test_rendering.py
```diff
@@ -202,6 +203,35 @@ def test_get_study_tables_cache(self):
...
+        GenericMaterial.objects.filter(
+            study=self.study, item_type='SOURCE'
+        ).update(characteristics=characteristics)
```
mikkonie, line 224: Minor point, but we are altering all columns now when just the first would be enough. (Did this test ISA-Tab even have multiple rows?)

samplesheets/tests/test_rendering.py
```diff
@@ -202,6 +203,35 @@ def test_get_study_tables_cache(self):
...
+
+        tables = self.tb.get_study_tables(self.study)
+        t_field = tables['study']['top_header'][0]
+        self.assertEqual(t_field['value'], 'Source')
```
mikkonie: Instead of asserting the top header value here, we should assert that the field header at index 2 is "age".

samplesheets/tests/test_rendering.py
```diff
@@ -202,6 +203,35 @@ def test_get_study_tables_cache(self):
...
+        val_field = tables['study']['table_data'][2]
+        self.assertEqual(val_field[2]['value'], '70')
```
mikkonie, line 228: Other than the minor comments above, good work with this test!

Now, make sure to also add another test where we do the same modification, but with the cache enabled, and ensure the cached value is returned instead of our modification.

samplesheets/tests/test_rendering.py
mikkonie: This new update is better, but the name is still confusing as we are not testing the sheet config at all here, just building the study tables. the sheet config is the auxillary configuration where we store settings related to editing the sheets.

Rename this into e.g. `test_get_study_tables_update_no_cache()`

samplesheets/tests/test_rendering.py
```diff
@@ -202,6 +203,35 @@ def test_get_study_tables_cache(self):
...
+        """Test get_study_tables() with SHEETS_ENABLE_STUDY_TABLE_CACHE=False"""
+        tables = self.tb.get_study_tables(self.study)
+        t_field = tables['study']['top_header'][0]
+        self.assertEqual(t_field['value'], 'Source')
+
+        sheet_config = self.build_sheet_config(self.investigation)
+        c_field = sheet_config['studies'][str(self.study.sodar_uuid)]['nodes'][
+            0
+        ]
+        self.assertEqual(c_field['header'], 'Source')
```
mikkonie: We can omit these lines related to the sheet config, as it is irrelevant for this test.

PR #1665, opened 2023-05-11
Irods delete request RESP API views

gromdimon:
### TODO:
- [x] Add Serializer
- [x] Add API views
- [x] Add view tests
- [x] Add permissions tests
- [x] Document

PR #1681, opened 2023-05-16
Irods delete request REST API views

gromdimon:
### TODO:
- [x] Add Serializer
- [x] Add API views
- [x] Add view tests
- [x] Add permissions tests
- [x] Document

### Previous PR: #1665


docs_manual/source/data_transfer_irods.rst
```diff
@@ -29,7 +29,7 @@ information.
 
 .. note::
 
-    On Ubuntu 20.04, installing iCommands is not officially supported at the
+    On Ubuntu 22.04, installing iCommands is not officially supported at the
```
mikkonie: The link below is in fact for a discussion about support for ubuntu 20.04. Nowadays there is support, so this entire section should be updated to reflect that 22.04 is not supported for the currently used iRODS version. However, since we'll be upgrading to iRODS 4.3 in this release, it can be removed entirely.

samplesheets/serializers.py
```diff
@@ -1,5 +1,8 @@
 """API view model serializers for the samplesheets app"""
 
+import re
+from django.conf import settings
```
mikkonie, line 5: Please follow the usual conventions for import ordering and annotation (I think I need to document these better in SODAR Core, come to think..)

Ordering:

- Imports of full built-in or non-django-specific modules (e.g. `import re`)
- Partial imports (e.g. `from datetime import date`)
- Django imports (e.g. `from django.conf import settings`)
- django-depending module imports (e.g. `from rest_framework import serializers`)
- `projectroles` imports
    * prefix with `# Projectroles dependency`
- Imports from within the current app

samplesheets/serializers.py
```diff
@@ -86,3 +90,69 @@ class Meta:
...
+
+    class Meta:
+        model = IrodsDataRequest
+        fields = ['path', 'description']
+
+    def validate_path(self, value):
+        irods_backend = get_backend_api('omics_irods')
+        # Remove trailing slashes as irodspython client does not recognize
+        # this as a collection
+        path = value.rstrip('/')
```
mikkonie: I recommend using `irods_backend.sanitize_path()` here. That also prevents potential parent string abuse.

samplesheets/serializers.py
```diff
@@ -86,3 +90,69 @@ class Meta:
...
+        ).first()
+        if old_request and old_request != self.instance:
+            raise serializers.ValidationError(ERROR_MSG_EXISTING)
+
+        path_re = re.compile(
+            '^' + irods_backend.get_projects_path() + '/[0-9a-f]{2}/'
+            '(?P<project_uuid>[0-9a-f-]{36})/'
+            + settings.IRODS_SAMPLE_COLL
+            + '/study_(?P<study_uuid>[0-9a-f-]{36})/'
+            'assay_(?P<assay_uuid>[0-9a-f-]{36})/.+$'
```
mikkonie, line 114: I recommend defining this regex once in `samplesheets.constants` and importing it to both `forms` and `serializers`.

samplesheets/views.py
```diff
@@ -771,6 +771,29 @@ def add_tl_create(cls, irods_request):
             name=irods_request.get_display_name(),
         )
 
+    @classmethod
+    def add_tl_update(cls, irods_request, timeline=None):
+        """
+        Create timeline event for iRODS data request update.
```
mikkonie, line 855: Missing empty line before params :) Also noticed the same for a couple of other methods, so please review.

mikkonie, line 855: Missing empty line before params :)

samplesheets/views_api.py
```diff
@@ -137,6 +146,268 @@ def post(self, request, *args, **kwargs):
         )
 
 
+class IrodsDataRequestListAPIView(
+    IrodsRequestModifyMixin, SODARAPIBaseProjectMixin, APIView
+):
+    """
+    List iRODS data requests for a project.
```
mikkonie, line 154: Make sure to add empty lines between the sections in API view docstrings, this will affect the rendering in the sphinx docs. Make sure to build the docs yourself and test to make sure they appear as expected.

samplesheets/views_api.py
```diff
@@ -137,6 +146,268 @@ def post(self, request, *args, **kwargs):
...
+            or project.is_delegate(self.request.user)
+            or project.is_owner(self.request.user)
+        ):
+            requests = irods_requests.filter(status__in=['ACTIVE', 'FAILED'])
+        else:
+            # For regular users, dispaly their own requests regardless of status
+            requests = irods_requests.filter(user=self.request.user)
+
+        if not requests:
+            return Response(
```
mikkonie: This can be simplified slightly by just setting up the content dict and then returning `Response()`.

samplesheets/views_api.py
```diff
@@ -52,6 +60,7 @@
 MD5_RE = re.compile(r'([a-fA-F\d]{32})')
 APP_NAME = 'samplesheets'
 IRODS_QUERY_ERROR_MSG = 'Exception querying iRODS objects'
+IRODS_EX_MSG = ' iRODS data request failed: '
```
mikkonie: Generally, I recommend not having spaces in the end/beginning of these types of partial strings. Instead, set up the correct spacing/punctiation in where you build the final message.

samplesheets/views_api.py
```diff
@@ -137,6 +146,268 @@ def post(self, request, *args, **kwargs):
...
+    permission_required = 'samplesheets.edit_sheet'
+
+    def post(self, request, *args, **kwargs):
+        """POST request for updating an iRODS data request"""
+        ex_msg = 'Updating' + IRODS_EX_MSG
+        irods_request = IrodsDataRequest.objects.filter(
+            sodar_uuid=self.kwargs.get('irodsdatarequest')
+        ).first()
+        if not irods_request:
+            raise ValidationError('{}Request not found'.format(ex_msg))
```
mikkonie: You probably want to add a `: ` or something here to clarify the final string. Same with other similar occurrences.

Furthermore, for views with `irodsdatarequest` as the URL object, we already fail in the permission check before we ever reach here, so this check is in fact not needed.

samplesheets/views.py
```diff
@@ -882,6 +905,169 @@ def handle_alerts_deactivate(cls, irods_request, app_alerts=None):
...
+    ):
+        """
+        Accept an iRODS accept request.
+        :param irods_request: IrodsDataRequest object
+        :param project: Project object
+        :param request: Request object
+        :param timeline: Timeline API or None
+        :param taskflow: Taskflow API or None
+        :param app_alerts: Appalerts API or None
+        :raise: ValueError if request is not in ACCEPTABLE_STATUSES
```
mikkonie: There are some references to `ACCEPTABLE_STATUSES` but I don't think this is defined or used anywhere?

samplesheets/views_api.py
```diff
@@ -137,6 +146,268 @@ def post(self, request, *args, **kwargs):
...
+    IrodsRequestModifyMixin, SODARAPIBaseProjectMixin, UpdateAPIView
+):
+    """
+    Update an iRODS data request for a project.
+    **URL:** ``/samplesheets/api/irods/request/update/{IrodsDataRequest.sodar_uuid}``
+    **Methods:** ``POST``
+    """
+
+    http_method_names = ['post']
+    permission_required = 'samplesheets.edit_sheet'
```
mikkonie: The permission to update or delete a request should only be granted to one of the following:

- Superuser
- Project owner or delegate (or: users with `manage_sheet` permission) 
- Creator of the request (e.g. `IrodsDataRequest.user`)

Please update this logic and permission tests.

samplesheets/views_api.py
```diff
@@ -137,6 +146,268 @@ def post(self, request, *args, **kwargs):
...
+    IrodsRequestModifyMixin, SODARAPIBaseProjectMixin, APIView
+):
+    """
+    Delete an iRODS data request for a project.
+    **URL:** ``/samplesheets/api/irods/request/delete/{IrodsDataRequest.sodar_uuid}``
+    **Methods:** ``DELETE``
+    """
+
+    http_method_names = ['delete']
+    permission_required = 'samplesheets.delete_sheet'
```
mikkonie: Same as with the update view, we should not let regular contributors delete each others' requests.

samplesheets/tests/test_permissions_api_taskflow.py
```diff
@@ -0,0 +1,361 @@
...
+    def test_create(self):
+        """Test post() in IrodsRequestCreateAPIView"""
+        good_users = [
+            self.superuser,
+            self.user_owner_cat,  # Inherited owner
+            self.user_owner,
+            self.user_delegate,
+            self.user_contributor,
+        ]
+        bad_users = [self.user_guest, self.user_no_roles, self.user_no_roles]
```
mikkonie: `user_no_roles` is duplicated here. This also happens elsewhere.

samplesheets/serializers.py
```diff
@@ -86,3 +92,70 @@ class Meta:
...
+
+    class Meta:
+        model = IrodsDataRequest
+        fields = ['path', 'description']
+
+    def validate_path(self, value):
+        irods_backend = get_backend_api('omics_irods')
+        # Remove trailing slashes as irodspython client does not recognize
+        # this as a collection
+        # path = value.rstrip('/')
```
mikkonie: Seems this line has been forgotten in comments :)

samplesheets/serializers.py
mikkonie: It seems the old comment is still valid.

samplesheets/serializers.py
```diff
@@ -86,3 +92,70 @@ class Meta:
...
+
+        old_request = IrodsDataRequest.objects.filter(
+            path=path, status__in=['ACTIVE', 'FAILED']
+        ).first()
+        if old_request and old_request != self.instance:
+            raise serializers.ValidationError(ERROR_MSG_EXISTING)
+
+        match = re.search(
+            path_re,
+            path,
```
mikkonie: Some wild unneeded line breaks appear! :)

samplesheets/tests/test_permissions_api_taskflow.py
```diff
@@ -74,3 +83,329 @@ def test_get_archive(self):
...
+            kwargs={'project': self.project.sodar_uuid},
+        )
+        # Set up post data
+        self.post_data = {'path': self.path + '/', 'description': 'bla'}
+
+    def test_create(self):
+        """Test post() in IrodsRequestCreateAPIView"""
+        good_users = [
+            self.superuser,
+            self.user_owner_cat,  # Inherited owner
```
mikkonie: Remember to add the new SODAR Core v0.13 users to all permission tests.

samplesheets/tests/test_views_api_taskflow.py
```diff
@@ -59,6 +69,87 @@ def setUp(self):
...
+class TestIrodsRequestAPIViewBase(
+    SampleSheetIOMixin, SampleSheetTaskflowMixin, TaskflowAPIViewTestBase
+):
+    """Base samplesheets API view test class for iRODS delete requests"""
+
+    def _assert_alert_count(self, alert_name, user, count, project=None):
+        """
+        Assert expected app alert count. If project is not specified, default to
+        self.project.
+        :param alert_name: String
```
mikkonie, line 81: Blank line missing before params.

samplesheets/views.py
```diff
@@ -885,6 +909,171 @@ def handle_alerts_deactivate(cls, irods_request, app_alerts=None):
...
+    ):
+        """
+        Reject an iRODS delete request.
+
+        :param irods_request: IrodsDataRequest object
+        :param project: Project object
+        :param request: Request object
+        :param timeline: Timeline API or None
+        :param app_alerts: Appalerts API or None
+        :raise: ValueError if request is not in ACCEPTABLE_STATUSES
```
mikkonie: We have a reference to `ACCEPTABLE_STATUSES` here but it seems not to exist anywhere.

samplesheets/views_api.py
```diff
@@ -52,6 +61,7 @@
 MD5_RE = re.compile(r'([a-fA-F\d]{32})')
 APP_NAME = 'samplesheets'
 IRODS_QUERY_ERROR_MSG = 'Exception querying iRODS objects'
+IRODS_EX_MSG = ' iRODS data request failed:'
```
mikkonie: Generally, I would recommend having trailing colons (or other trailing formatting characters) in these kinds of potentially reusable string constants. Instead, add these when building the final string.

samplesheets/views_api.py
```diff
@@ -137,6 +147,317 @@ def post(self, request, *args, **kwargs):
...
+        project = Project.objects.filter(
+            sodar_uuid=self.kwargs.get('project')
+        ).first()
+        irods_requests = IrodsDataRequest.objects.filter(project=project)
+
+        # For superusers, owners and delegates,
+        # display active/failed requests from all users
+        if (
+            self.request.user.is_superuser
+            or project.is_delegate(self.request.user)
```
mikkonie: Hint: You can also use `is_owner_or_delegate()` here :)

samplesheets/views_api.py
```diff
@@ -137,6 +147,317 @@ def post(self, request, *args, **kwargs):
...
+        """Override to allow POST requests"""
+        return IrodsDataRequest.objects.get(
+            sodar_uuid=self.kwargs.get('irodsdatarequest')
+        )
+
+    def check_irods_permissions(self, request):
+        """Override to allow POST requests without CSRF token"""
+        if (
+            request.user.is_superuser
+            or self.get_project().is_owner(request.user)
```
mikkonie: We can omit the `is_owner()` check here, as owner already has `manage_sheet` perms.

samplesheets/views_api.py
```diff
@@ -137,6 +147,317 @@ def post(self, request, *args, **kwargs):
...
+            or request.user.has_perm(
+                'samplesheets.manage_sheet', self.get_object().project
+            )
+            or request.user == self.get_object().user
```
mikkonie: We can just do `project = self.get_object()` in the beginning of the method and potentially omit one redundant database query :)

samplesheets/views_api.py
```diff
@@ -137,6 +147,317 @@ def post(self, request, *args, **kwargs):
...
+        except Exception as ex:
+            raise APIException('{} {}'.format(ex_msg, ex))
+
+        # Add timeline event
+        self.add_tl_delete(irods_request)
+        # Handle project alerts
+        self.handle_alerts_deactivate(irods_request)
+        return Response(
+            {
+                'detail': 'iRODS data request deleted',
```
mikkonie: I think here we can again omit some unneeded line breaks by removing the comma and placing the dict on one line.

samplesheets/tests/test_permissions_api_taskflow.py
```diff
@@ -74,3 +83,329 @@ def test_get_archive(self):
...
+
+    def test_update(self):
+        """Test post() in IrodsRequestUpdateAPIView"""
+        good_users = [
+            self.superuser,
+            self.user_owner_cat,  # Inherited owner
+            self.user_owner,
+            self.user_delegate,
+        ]
+        bad_users = [self.user_guest, self.user_no_roles, self.user_contributor]
```
mikkonie: Let's maintain a consistent ordering of users so it's easier to keep track of these tests. Contributor should come before guest.

samplesheets/tests/test_permissions_api_taskflow.py
```diff
@@ -74,3 +83,329 @@ def test_get_archive(self):
...
+            self.user_contributor,
+        ]
+        bad_users = [self.user_guest, self.user_no_roles]
+
+        self.assert_response_api(
+            self.url, good_users, 200, method='POST', data=self.post_data
+        )
+        self.assert_response_api(
+            self.url, bad_users, 403, method='POST', data=self.post_data
+        )
```
mikkonie, line 157: Also rememeber to test with `anonymous` (should get a 401)

samplesheets/tests/test_permissions_api_taskflow.py
```diff
@@ -74,3 +83,329 @@ def test_get_archive(self):
...
+    def test_create_anon(self):
+        """Test post() in IrodsRequestCreateAPIView with anonymous access"""
+        self.project.set_public()
+        self.assert_response_api(
+            self.url,
+            self.anonymous,
+            401,
+            method='POST',
+            data=self.post_data,
+        )
```
mikkonie, line 172: In addition to anonymous access, make sure to also test permissions if project is set as archived.

samplesheets/tests/test_permissions_api_taskflow.py
```diff
@@ -74,3 +83,329 @@ def test_get_archive(self):
...
+        good_users = [
+            self.superuser,
+            self.user_owner_cat,  # Inherited owner
+            self.user_owner,
+            self.user_delegate,
+        ]
+        bad_users = [self.user_guest, self.user_no_roles, self.user_contributor]
+
+        for user in good_users:
+            with self.login(self.superuser):
```
mikkonie: Instead of this, we can set up a method to create a request, call it once before the tests and then pass it as `cleanup_method` to `assert_response_api()`. It's a bit less compact way of doing this. See for examples elsewhere in SODAR and SODAR Core.

samplesheets/tests/test_permissions_api_taskflow.py
```diff
@@ -74,3 +83,329 @@ def test_get_archive(self):
...
+            self.user_owner,
+            self.user_delegate,
+        ]
+        bad_users = [self.user_contributor, self.user_guest, self.user_no_roles]
+        with self.login(self.superuser):
+            self.client.post(self.url_create, self.post_data)
+            obj = IrodsDataRequest.objects.first()
+            self.assertIsNotNone(obj)
+
+        self.url_accept = reverse(
```
mikkonie, line 362: Hmm, it seems here you are re-accepting the same request multiple times. This should not be allowed, see my comment in `accept_request()`. Once that method has been changed, use `cleanup_method` to set up a new request to be accepted between iterations with different users.

samplesheets/views.py
```diff
@@ -885,6 +909,171 @@ def handle_alerts_deactivate(cls, irods_request, app_alerts=None):
...
+
+        :param irods_request: IrodsDataRequest object
+        :param project: Project object
+        :param request: Request object
+        :param timeline: TimelineAPI instance or None
+        :param taskflow: TaskflowAPI instance or None
+        :param app_alerts: AppalertsAPI instance or None
+        :raise: FlowSubmitException if taskflow submission fails
+        """
+        tl_event = None
```
mikkonie, line 1009: We should check and fail if the status of the request is already `ACCEPTED`. Accepting the same request multiple times doesn't make sense. This might happen if different users open the accept form in the UI or if the accept API view is called multiple times.

This should raise an exception and also update the timeline event with `FAILED` status and an appropriate description.

And yes, this seems to have been missing from the original implementation. It's a bigger deal now that we have the API views.

samplesheets/tests/test_views_api_taskflow.py
```diff
@@ -144,6 +235,644 @@ def test_post_created(self):
...
+
+        obj.refresh_from_db()
+        self.assertEqual(obj.status, 'FAILED')
+        self.assert_irods_obj(self.path, True)
+        self._assert_alert_count(ACCEPT_ALERT, self.user, 0)
+        self._assert_alert_count(ACCEPT_ALERT, self.user_delegate, 0)
+        self._assert_alert_count(ACCEPT_ALERT, self.user_contrib, 0)
+        self._assert_alert_count(CREATE_ALERT, self.user, 1)
+        self._assert_alert_count(CREATE_ALERT, self.user_delegate, 1)
+        self.assert_irods_obj(self.path, True)
```
mikkonie, line 672: Add a test for accepting an already accepted request. This should fail, see comment in `accept_request()`.

samplesheets/tests/test_permissions_api_taskflow.py
```diff
@@ -74,3 +83,329 @@ def test_get_archive(self):
...
+        good_users = [
+            self.superuser,
+            self.user_owner_cat,  # Inherited owner
+            self.user_owner,
+            self.user_delegate,
+        ]
+        bad_users = [self.user_contributor, self.user_guest, self.user_no_roles]
+
+        with self.login(self.superuser):
+            self.client.post(self.url_create, self.post_data)
```
mikkonie: Same thing as I mentioned elsewhere, I recommend using `cleanup_method`.

samplesheets/tests/test_views_api_taskflow.py
```diff
@@ -19,11 +20,20 @@
 )
 
 # Samplesheets dependencies
+from samplesheets.views import (
+    IRODS_REQ_CREATE_ALERT as CREATE_ALERT,
+    IRODS_REQ_ACCEPT_ALERT as ACCEPT_ALERT,
+    IRODS_REQ_REJECT_ALERT as REJECT_ALERT,
+)
 from samplesheets.views_api import IRODS_QUERY_ERROR_MSG
+from samplesheets.models import IrodsDataRequest
```
mikkonie: Super minor point, but I prefer placing these imports in alphabetical order, so `samplesheets.models` should come before the view imports.

samplesheets/tests/test_views_api_taskflow.py
```diff
@@ -144,6 +235,644 @@ def test_post_created(self):
...
+        with self.login(self.user_contrib):
+            response = self.client.post(url, self.post_data)
+            self.assertEqual(response.status_code, 200)
+
+        obj = IrodsDataRequest.objects.first()
+        self.assertEqual(IrodsDataRequest.objects.count(), 1)
+        self.assertEqual(obj.path, self.path)
+        self.assertEqual(obj.description, 'bla')
+        self._assert_alert_count(CREATE_ALERT, self.user, 1)
+        self._assert_alert_count(CREATE_ALERT, self.user_delegate, 1)
```
mikkonie: Maybe also check here that a contributor/guest did not receive any alert? Oh and as a general note, good work with these tests!

samplesheets/views_api.py
```diff
@@ -137,6 +147,317 @@ def post(self, request, *args, **kwargs):
...
+
+class IrodsRequestUpdateAPIView(
+    IrodsRequestModifyMixin, SODARAPIBaseProjectMixin, UpdateAPIView
+):
+    """
+    Update an iRODS data request for a project.
+
+    **URL:** ``/samplesheets/api/irods/request/update/{IrodsDataRequest.sodar_uuid}``
+
+    **Methods:** ``POST``
```
mikkonie: I think we should use `PUT` and `PATCH` in an update view, no?

samplesheets/tests/test_views_api_taskflow.py
```diff
@@ -144,6 +235,644 @@ def test_post_created(self):
...
+
+class TestIrodsRequestUpdateAPIView(TestIrodsRequestAPIViewBase):
+    """Tests for IrodsRequestUpdateAPIView"""
+
+    def test_update(self):
+        """Test updating a request"""
+        update_data = {'path': self.path, 'description': 'Updated'}
+
+        with self.login(self.user_contrib):
+            self.client.post(
```
mikkonie, line 333: Same comment as in the view class itself, I think we should use PUT and possibly PATCH in this update view instead of POST, to keep things consistent.

samplesheets/views_api.py
```diff
@@ -137,6 +147,317 @@ def post(self, request, *args, **kwargs):
...
+    """
+
+    http_method_names = ['post']
+    # permission_required = 'samplesheets.edit_sheet'
+    permission_classes = (IsAuthenticated,)
+
+    def check_irods_permissions(self, request):
```
mikkonie: This method seems to be duplicated in `IrodsRequestDeleteAPIView` and in both cases it's only called once within the view. I suggest to remove the repetition by making this a mixin.

The docstring is a bit confusing, as this is not an override of anything and the API views should have nothing to do with CSRF tokens. Maybe a forgotten copypaste from somewhere? :)

samplesheets/views_api.py
```diff
@@ -137,6 +147,317 @@ def post(self, request, *args, **kwargs):
...
+        """Override to allow POST requests without CSRF token"""
+        if (
+            request.user.is_superuser
+            or self.get_project().is_owner(request.user)
```
mikkonie: Owner is included in the `manage_sheet` perm check, so this separate check can be omitted.

samplesheets/serializers.py
gromdimon: Yep,sorry I didn't mention that. If I define this path in `samplesheets.constants` the following Error is being raised:
```
django.core.exceptions.AppRegistryNotReady: Models aren't loaded yet.
```
I think that is impossible due to irods_backend. An alternative way could be an f-string. What do you think on that?

samplesheets/tests/test_permissions_api_taskflow.py
gromdimon: The thing is, that I need a uuid of a created request, but cleanup_method can not provide it(

samplesheets/views_api.py
mikkonie: This comment is still valid in the latest version.

samplesheets/views_api.py
```diff
@@ -137,6 +147,288 @@ def post(self, request, *args, **kwargs):
...
+            sodar_uuid=self.kwargs.get('irodsdatarequest')
+        )
+
+    def delete(self, request, *args, **kwargs):
+        """DELETE request for deleting an iRODS data request"""
+        obj = self.get_object()
+        if not self.check_irods_permissions(request, obj):
+            raise PermissionDenied('Insufficient permissions')
+
+        ex_msg = 'Deleting' + IRODS_EX_MSG + ':'
```
mikkonie: We only have use for this if `delete()` raises an exception, hence I would simply form this message on one line when you do `raise APIException()`.

samplesheets/serializers.py
mikkonie: Ah right, that is a good point. Let's leave it like this for now, maybe later we can e.g. add some kind of helper in irodsbackend to generate this.

samplesheets/tests/test_permissions_api_taskflow.py
```diff
@@ -1,19 +1,28 @@
...
 
+from django.test import override_settings
 from django.urls import reverse
 
+from samplesheets.models import IrodsDataRequest
+
 # Taskflowbackend dependency
 from taskflowbackend.tests.base import TaskflowAPIPermissionTestBase
 
+# Samplesheets dependency
```
mikkonie: Since this is the samplesheets app, there's no need to write this :)

samplesheets/tests/test_permissions_api_taskflow.py
```diff
@@ -1,19 +1,28 @@
...
+"""Tests for samplesheets REST API view permissions with taskflow"""
+
+import os
 
 from irods.keywords import REG_CHKSUM_KW
 
+from django.test import override_settings
 from django.urls import reverse
 
+from samplesheets.models import IrodsDataRequest
```
mikkonie: Please move this together with the other samplesheets dependencies.

samplesheets/tests/test_permissions_api_taskflow.py
mikkonie: Still valid, see my comment about setting the UUID for `cleanup_method` in the other comment,

samplesheets/tests/test_permissions_api_taskflow.py
```diff
@@ -74,3 +83,493 @@ def test_get_archive(self):
...
+        )
+
+    def test_create_archived(self):
+        """Test post() in IrodsRequestCreateAPIView with archived project"""
+        self.project.set_archive()
+        good_users = [self.superuser]
+        bad_users = [
+            self.user_owner_cat,  # Inherited owner
+            self.user_owner,
+            self.user_delegate_cat,
```
mikkonie, line 180: Please update the ordering of users as I specified in the landing zone PR.

samplesheets/views_api.py
```diff
@@ -137,6 +147,288 @@ def post(self, request, *args, **kwargs):
...
+
+    **Methods:** ``GET``
+    """
+
+    http_method_names = ['get']
+    permission_required = 'samplesheets.manage_sheet'
+
+    def get(self, request, *args, **kwargs):
+        """POST request for rejecting an iRODS data request"""
+        ex_msg = 'Rejecting' + IRODS_EX_MSG + ':'
```
mikkonie: Here we can also just form the message when raising `APIException`, since it's only used if an exception is met.

samplesheets/views_api.py
```diff
@@ -137,6 +147,288 @@ def post(self, request, *args, **kwargs):
...
+
+    **Methods:** ``GET``
+    """
+
+    http_method_names = ['get']
+    permission_required = 'samplesheets.edit_sheet'
+
+    def get(self, request, *args, **kwargs):
+        """GET request for listing iRODS data requests"""
+        project = Project.objects.filter(
```
mikkonie: We can just use `self.get_project()` here.

samplesheets/views.py
```diff
@@ -885,6 +909,190 @@ def handle_alerts_deactivate(cls, irods_request, app_alerts=None):
...
+                    'samplesheets:project_sheets',
+                    kwargs={'project': project.sodar_uuid},
+                ),
+                project=project,
+            )
+            # Handle project alerts
+            cls.handle_alerts_deactivate(irods_request, app_alerts)
+
+    def check_irods_permissions(self, request, obj):
+        """Check permissions to a project"""
```
mikkonie: This docstring is vague. We should specify it as e.g. "Check permissions for a landing zone". Also, the `obj` should specifically be named as e.g. `zone` as this check is specifically for a landing zone.

Come to think of it, the method itself is named incorrectly. This has nothing to do with irods, only for modifying landing zones. Hence, e.g. `get_zone_permissions()` would be more appropriate.

samplesheets/tests/test_permissions_api_taskflow.py
mikkonie: Not a problem :) Create UUID at the start of test or in the `setUp()` method, then set that UUID to the newly created object in the cleanup method. Ta-da, object with a static UUID :)

samplesheets/views.py
```diff
@@ -885,6 +909,190 @@ def handle_alerts_deactivate(cls, irods_request, app_alerts=None):
...
+        :param taskflow: TaskflowAPI instance or None
+        :param app_alerts: AppalertsAPI instance or None
+        :raise: FlowSubmitException if taskflow submission fails
+        """
+        tl_event = None
+        description = 'accept iRODS data request {irods_request}'
+        status = 'OK'
+        if irods_request.status == 'ACCEPTED':
+            description = (
+                'iRODS data request {irods_request} is already ' 'accepted'
```
mikkonie: Looks like an unnecessarily split string here.

samplesheets/views.py
```diff
@@ -885,6 +909,190 @@ def handle_alerts_deactivate(cls, irods_request, app_alerts=None):
...
+        :param project: Project object
+        :param request: Request object
+        :param timeline: TimelineAPI instance or None
+        :param taskflow: TaskflowAPI instance or None
+        :param app_alerts: AppalertsAPI instance or None
+        :raise: FlowSubmitException if taskflow submission fails
+        """
+        tl_event = None
+        description = 'accept iRODS data request {irods_request}'
+        status = 'OK'
```
mikkonie: These two lines should go under `else` after the `ACCEPTED` check, or the other way around.

samplesheets/tests/test_views_api_taskflow.py
mikkonie: This still seems to be missing from the current version, or maybe I'm not seeing it?

samplesheets/tests/test_views_api_taskflow.py
```diff
@@ -19,11 +20,20 @@
 )
 
 # Samplesheets dependencies
```
mikkonie: This can also be removed as this module is in the samplesheets app. I guess it was there by mistake :)

samplesheets/views_api.py
```diff
@@ -52,6 +61,7 @@
 MD5_RE = re.compile(r'([a-fA-F\d]{32})')
 APP_NAME = 'samplesheets'
 IRODS_QUERY_ERROR_MSG = 'Exception querying iRODS objects'
+IRODS_EX_MSG = ' iRODS data request failed'
```
mikkonie: Same thing as with the trailing special characters, we should not have leading spaces or special characters in a reusable string constant.

samplesheets/tests/test_permissions_api_taskflow.py
gromdimon: That sounds quite good, but:
When I create a url object and then call the `assert_response_api()` method, then this url is being chnaged in cleanup_method. But the url, which is used for sending the requests (a local variable inside function) can not be changed that way (as in that case the whole function `assert_response_api()` have to be called once again, or somehow the local url variable changed to the url object, that I am creating in the beginning). 
So I spent quite time on that, but didn't find any proper way with cleanup_method(), which was better then a loop. Maybe I misunderstood you?

gromdimon:
### Hint:
The "changes" push crashes in CI, as it implements the `cleanup_methods` in tests.

samplesheets/tests/test_permissions_api_taskflow.py
mikkonie: I think my earlier explanation was not the best. You can see a recent example in this class, works fine :) See how `_make_ticket()` is set up and used. We simply pass the UUID of `self.ticket` in the tests.

https://github.com/bihealth/sodar-server/blob/dev/samplesheets/tests/test_permissions.py#L1178

mikkonie: We had a talk about this and I was wrong, in this case a built-in restriction in the SODAR Core implementation of this test feature does in fact prevent doing this. This will be improved in bihealth/sodar-core#1234.

samplesheets/views_api.py
mikkonie: ..and it still hasn't been changed :)

samplesheets/tests/test_views_api_taskflow.py
mikkonie: I'm still not seeing that test, is there a reason why this hasn't been implemented?

samplesheets/views.py
```diff
@@ -885,6 +909,191 @@ def handle_alerts_deactivate(cls, irods_request, app_alerts=None):
...
+                url=reverse(
+                    'samplesheets:project_sheets',
+                    kwargs={'project': project.sodar_uuid},
+                ),
+                project=project,
+            )
+            # Handle project alerts
+            cls.handle_alerts_deactivate(irods_request, app_alerts)
+
+    def get_zone_permissions(self, request, zone):
```
mikkonie: I think I was previously confused by two different pull requests :D

Of course this has nothing to do with any landing zones at all, but with irods data requests. Hence this should be called something like  `has_irods_request_perms()` and the object is `irods_request`, not `zone`.

Sorry for being confusing. I need a vacation..

Eventually, we should probably refactor this so that the whole check happens in `samplesheets.rules` instead of a hardcoded hack, but we can do that in a further update. If it works it works.

samplesheets/tests/test_permissions_api_taskflow.py
```diff
@@ -74,3 +81,493 @@ def test_get_archive(self):
...
+        # Set up iRODS data
+        self.make_irods_colls(self.investigation)
+        self.assay_path = self.irods_backend.get_path(self.assay)
+        self.path = os.path.join(self.assay_path, TEST_FILE_NAME)
+        self.path_md5 = os.path.join(self.assay_path, f'{TEST_FILE_NAME}.md5')
+        # Create objects
+        self.file_obj = self.irods.data_objects.create(self.path)
+        self.md5_obj = self.irods.data_objects.create(self.path_md5)
+
+    def create_request(self):
```
mikkonie: Super minor point: I prefer having helper methods before `setUp()`, so that `setUp()` (and possible `tearDown()`) always comes last in the base/helper methods.

samplesheets/tests/test_permissions_api_taskflow.py
```diff
@@ -74,3 +81,493 @@ def test_get_archive(self):
...
+            self.assert_response_api(
+                self.url_accept,
+                user,
+                200,
+                method='POST',
+                data={'confirm': True},
+            )
+
+        for user in bad_users:
+            obj = self.create_request()
```
mikkonie: Since the object is not actually changed for bad users, it should be enough to create the request just once before this loop, right? The same goes for other similar tests.

samplesheets/tests/test_views_api_taskflow.py
```diff
@@ -59,6 +68,88 @@ def setUp(self):
         self.assay = self.study.assays.first()
 
 
+class TestIrodsRequestAPIViewBase(
+    SampleSheetIOMixin, SampleSheetTaskflowMixin, TaskflowAPIViewTestBase
+):
+    """Base samplesheets API view test class for iRODS delete requests"""
+
+    def _assert_alert_count(self, alert_name, user, count, project=None):
```
mikkonie: Since it's a base view, we should not have this method protected, it can just be called `assert_alert_count()`. I know, I've done this myself in the past, but I'm working on correcting those :)

samplesheets/tests/test_views_api_taskflow.py
```diff
@@ -144,6 +235,646 @@ def test_post_created(self):
...
 
+class TestIrodsRequestCreateAPIView(TestIrodsRequestAPIViewBase):
+    """Tests for IrodsRequestCreateAPIView"""
+
+    def test_create(self):
+        """Test post() in IrodsRequestCreateAPIView"""
+        self.assertEqual(IrodsDataRequest.objects.count(), 0)
+        self._assert_alert_count(CREATE_ALERT, self.user, 0)
+        self._assert_alert_count(CREATE_ALERT, self.user_delegate, 0)
+        url = reverse(
```
mikkonie: If the URL for the requests is always the same within a test class, we can make these tests a bit more concise by defining it as `self.url` in `setUp()`.

I've just recently started doing this myself, as some test modules are startting to grow rather big (as you probably have noticed :))

samplesheets/tests/test_views_api_taskflow.py
```diff
@@ -144,6 +235,646 @@ def test_post_created(self):
...
+            self.client.post(
+                reverse(
+                    'samplesheets:api_irods_request_create',
+                    kwargs={'project': self.project.sodar_uuid},
+                ),
+                self.post_data,
+            )
+            self.assertEqual(IrodsDataRequest.objects.count(), 1)
+            self._assert_alert_count(CREATE_ALERT, self.user, 1)
+            self._assert_alert_count(CREATE_ALERT, self.user_delegate, 1)
```
mikkonie: In these deletion tests we probably aren't that interested in request creation alerts. Also, as a minor point, after `with self.login()..` you don't have to indent anything except for redirect asserts (those still require the user to be logged in).

samplesheets/tests/test_views_api_taskflow.py
```diff
@@ -144,6 +235,646 @@ def test_post_created(self):
...
+                reverse(
+                    'samplesheets:api_irods_request_create',
+                    kwargs={'project': self.project.sodar_uuid},
+                ),
+                self.post_data,
+            )
+
+        self.assertEqual(IrodsDataRequest.objects.count(), 1)
+        obj = IrodsDataRequest.objects.first()
+        self._assert_alert_count(CREATE_ALERT, self.user, 1)
```
mikkonie: the create alerts are not that critical to test in this class, we already establish they are created in the create view test class.

samplesheets/tests/test_views_api_taskflow.py
```diff
@@ -144,6 +235,646 @@ def test_post_created(self):
...
+        self._assert_alert_count(CREATE_ALERT, self.user_delegate, 1)
+        self._assert_alert_count(ACCEPT_ALERT, self.user, 0)
+        self._assert_alert_count(ACCEPT_ALERT, self.user_delegate, 0)
```
mikkonie: Maytbe add an assert for `ACCEPT_ALERT` and `user_contrib` here?

samplesheets/views_api.py
```diff
@@ -137,6 +147,289 @@ def post(self, request, *args, **kwargs):
...
+                app_alerts=app_alerts,
+            )
+        except Exception as ex:
+            raise APIException(
+                '{} {}'.format('Rejecting ' + IRODS_EX_MSG + ':', ex)
+            )
+
+        return Response(
+            {
+                'detail': 'iRODS data request rejected',
```
mikkonie: This is unnecessarily split into multiple lines.

samplesheets/tests/test_views_api_taskflow.py
```diff
@@ -144,6 +240,645 @@ def test_post_created(self):
...
+
+        with self.login(self.user):
+            response = self.client.post(
+                reverse(
+                    'samplesheets:api_irods_request_accept',
+                    kwargs={'irodsdatarequest': obj.sodar_uuid},
+                ),
+                {'confirm': True},
+            )
+        self.assertEqual(response.status_code, 500)
```
mikkonie: Good work with the test. However, this should return 400 (bad request), not 500 (internal server error).

samplesheets/tests/test_views_api_taskflow.py
```diff
@@ -144,6 +240,645 @@ def test_post_created(self):
...
+            response = self.client.delete(
+                reverse(
+                    'samplesheets:api_irods_request_delete',
+                    kwargs={'irodsdatarequest': obj.sodar_uuid},
+                )
+            )
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(IrodsDataRequest.objects.count(), 1)
+        self.assert_alert_count(CREATE_ALERT, self.user, 1)
```
mikkonie: We are still looking for `CREATE_ALERT` alerts here for a delete method. Probably not that relevant here? If we create alerts about request deletion (I'm not sure if we do?) then those should be tested instead.

docs_manual/source/api_samplesheets.rst
```diff
@@ -23,6 +35,8 @@ API Views
 
 .. autoclass:: ProjectIrodsFileListAPIView
 
+.. autoclass:: RemoteSheetGetAPIView
```
mikkonie: This is a legacy hack view which will be removed. It's been kept out of the documentation for a reason, to discourage anyone from using it :)

samplesheets/forms.py
```diff
@@ -402,6 +402,7 @@ def clean(self):
             + '/study_(?P<study_uuid>[0-9a-f-]{36})/'
             'assay_(?P<assay_uuid>[0-9a-f-]{36})/.+$'
         )
+
```
mikkonie: There seems to be an unnecessary addition of an empty line here.

samplesheets/tests/test_permissions_api_taskflow.py
```diff
@@ -74,3 +81,489 @@ def test_get_archive(self):
         ]
         self.assert_response_api(url, good_users, 200, data=self.post_data)
         self.assert_response_api(url, self.anonymous, 401, data=self.post_data)
+
+
+class TestIrodsRequestAPIViewBase(
+    SampleSheetIOMixin, SampleSheetTaskflowMixin, TaskflowAPIPermissionTestBase
+):
+    """Helper mixin for IrodsRequestAPIView tests"""
```
mikkonie: This is a base class, not a mixin :)

samplesheets/tests/test_views_api_taskflow.py
```diff
@@ -8,6 +8,7 @@
 from irods.keywords import REG_CHKSUM_KW
 
 from django.urls import reverse
+from django.test import override_settings
```
mikkonie, line 10: Super minor note, but let's place dependencies from the same module into alphabetical order so they're easier to track and edit.

samplesheets/views.py
```diff
@@ -961,6 +985,193 @@ def handle_alerts_deactivate(cls, irods_request, app_alerts=None):
...
+        cls,
+        irods_request,
+        project,
+        request,
+        timeline=None,
+        taskflow=None,
+        app_alerts=None,
+    ):
+        """
+        Accept an iRODS accept request.
```
mikkonie: This should probably say "iRODS data request" :)

mikkonie (changes requested):

samplesheets/tests/test_views_api_taskflow.py
gromdimon: Just missed this pair :(

samplesheets/views_api.py
```diff
@@ -137,6 +147,275 @@ def post(self, request, *args, **kwargs):
...
+class IrodsDataRequestListAPIView(
+    IrodsRequestModifyMixin, SODARAPIBaseProjectMixin, APIView
+):
+    """
+    List iRODS data requests for a project.
+
+    **URL:** ``/samplesheets/api/irods/requests/{Project.sodar_uuid}``
+
+    **Methods:** ``GET``
+    """
```
mikkonie, line 163: The `Returns:` information is missing from the API view docstring. For examples, see e.g. [projectroles REST API docs](https://sodar-core.readthedocs.io/en/dev/app_projectroles_api_rest.html)

samplesheets/views_api.py
```diff
@@ -137,6 +147,275 @@ def post(self, request, *args, **kwargs):
...
+class IrodsRequestAcceptAPIView(
+    IrodsRequestModifyMixin, SODARAPIBaseProjectMixin, APIView
+):
+    """
+    Accept an iRODS data request for a project.
+
+    **URL:** ``/samplesheets/api/irods/request/accept/{IrodsDataRequest.sodar_uuid}``
+
+    **Methods:** ``POST``
+    """
```
mikkonie, line 341: `Returns` info also missing here, since we return the `path` in addition to `detail`.

I'm not sure if we even need to return the path, as the requesting party should know it already if they are accepting the request. Instead of updating the docstring, the field can probably just be omitted :)

samplesheets/views_api.py
```diff
@@ -137,6 +147,275 @@ def post(self, request, *args, **kwargs):
...
+                project,
+                self.request,
+                timeline=timeline,
+                app_alerts=app_alerts,
+            )
+        except Exception as ex:
+            raise APIException(
+                '{} {}'.format('Rejecting ' + IRODS_EX_MSG + ':', ex)
+            )
+
```
mikkonie: Please remove redundant empty line.

samplesheets/tests/test_permissions_api_taskflow.py
```diff
@@ -74,3 +81,489 @@ def test_get_archive(self):
...
+        bad_users = [
+            self.user_contributor_cat,
+            self.user_guest_cat,
+            self.user_finder_cat,
+            self.user_contributor,
+            self.user_guest,
+            self.user_no_roles,
+        ]
+
+        obj = self.create_request()
```
mikkonie: I believe we only need to create the object once in `setUp()` for testing the permissions of an update view. Even if we update the model repeatedly with the same values, the response codes should still be the same.

The URL can also be set once in `self.url` inside `setUp()` in this case.

samplesheets/tests/test_permissions_api_taskflow.py
```diff
@@ -74,3 +81,489 @@ def test_get_archive(self):
         ]
         self.assert_response_api(url, good_users, 200, data=self.post_data)
         self.assert_response_api(url, self.anonymous, 401, data=self.post_data)
+
+
+class TestIrodsRequestAPIViewBase(
+    SampleSheetIOMixin, SampleSheetTaskflowMixin, TaskflowAPIPermissionTestBase
+):
+    """Base test class for IrodsRequestAPIView tests"""
```
mikkonie: This should probably read `Base test class for IrodsRequestAPIView permission tests`.

samplesheets/tests/test_permissions_api_taskflow.py
```diff
@@ -74,3 +81,489 @@ def test_get_archive(self):
...
+            self.url_accept = reverse(
+                'samplesheets:api_irods_request_accept',
+                kwargs={'irodsdatarequest': obj.sodar_uuid},
+            )
+            self.assert_response_api(
+                self.url_accept,
+                user,
+                200,
+                method='POST',
+                data={'confirm': True},
```
mikkonie: Where does this `confirm` parameter come from? The view should not require that and I don't see it included anywhere.

samplesheets/tests/test_views_api_taskflow.py
```diff
@@ -144,6 +240,643 @@ def test_post_created(self):
...
+        self.assert_alert_count(ACCEPT_ALERT, self.user_delegate, 0)
+        self.assert_alert_count(ACCEPT_ALERT, self.user_contrib, 0)
+
+        with self.login(self.user):
+            response = self.client.post(
+                reverse(
+                    'samplesheets:api_irods_request_accept',
+                    kwargs={'irodsdatarequest': obj.sodar_uuid},
+                ),
+                {'confirm': True},
```
mikkonie: Here I'm also not sure why the `confirm` parameter is used. That should not be required.

mikkonie (changes requested):
We are 99% there, but I still noticed some minor things.

samplesheets/tests/test_permissions_api_taskflow.py
gromdimon: This `confirm=True` parameter is that small checkbox under the list of affected objects on the confirmotion page. We can ommit it, but in theory it will cause `form.is_valid = False` (what in best case just will return the same page with filled form) and it is not what we want I am afraid(

gromdimon: Oh, sorry, I was messed up by UI views, where there's a checkbox for confirmation(

samplesheets/tests/test_permissions_api_taskflow.py
```diff
@@ -74,3 +81,478 @@ def test_get_archive(self):
...
+            self.url_accept = reverse(
+                'samplesheets:api_irods_request_accept',
+                kwargs={'irodsdatarequest': obj.sodar_uuid},
+            )
+            self.assert_response_api(
+                self.url_accept,
+                user,
+                200,
+                method='POST',
+                # data={'confirm': True},
```
mikkonie: Please remove the commented out `data={'confirm': True}` lines :)

mikkonie (changes requested):

mikkonie (approved):
Looks great, will be merged :)

PR #1682, opened 2023-05-16
Add davrods link to iRODS deletion request list items & Implement iRODS accept / reject batch update

### LINKED ISSUE #1340: Add batch accepting/rejecting for iRODS delete requests
I discussed this with @holtgrewe and it would make sense to allow batch accepting or rejecting of iRODS delete requests, with some fallbacks.

## Spec

- Add checkboxes for delete requests
- Replace `Create Request` button with a `Request Operations` dropdown with the following options:
    * Create Request
    * Accept Selected
    * Reject Selected
- Make "Accept Selected" and "Reject Selected" options disabled, if no checkboxes are selected
    * This has to be done in JQuery of course
- Modify `IrodsRequestAcceptView` and `IrodsRequestRejecttView` to accept a list of objects
    * I believe `IrodsRequestAcceptView` should be changed from `FormView` to `View` similar to `IrodsRequestRejecttView`.. But whatever works is ok :)
- Make sure to list all files to be deleted from each request in the accept confirmation view
- Add tests
- Update docs to note that accepting/rejecting multiple requests is possible this way
- **HINT:** Check filesfolders in SODAR Core for an example how we have done this before
    * ..at least regarding the UI & JQuery side
    * The batch delete/move view itself is quite messy and is not probably the best example :)

gromdimon:
### TODO:
- [x] In IrodsDataRequestListView, add Davrods link button for each item
- [x] Add as button next to the path clipboard copy button
- [x] Test for existence in TestIrodsDataRequestListView UI tests

### TODO:
- [x] Add checkboxes for delete requests
- [x] Replace Create Request button with a Request Operations dropdown
- [x] Make "Accept Selected" and "Reject Selected" options disabled, if no checkboxes are selected
- [x] Modify IrodsRequestAcceptView and IrodsRequestRejecttView to accept a list of objects
- [x] Add tests
- [x] Update docs to note that accepting/rejecting multiple requests is possible this way


samplesheets/templates/samplesheets/irods_requests.html
```diff
@@ -153,6 +160,24 @@ <h3>iRODS Delete Requests</h3>
                   {% endif %}
                 </div>
               </td>
+              {% if irods_webdav_enabled %}
+                  <td class="text-right" id="sodar-ss-davrods-link">
+                    <a href="{{ irods_request.webdav_url }}"
```
mikkonie: As specified in the original spec, the button should go next to the clipboard copying button on the right side of the 1st column in the table.

samplesheets/static/samplesheets/js/irods_request.js
```diff
@@ -0,0 +1,80 @@
+// JS for iRODS request page
+$(document).ready(function () {
+    // Disable "Accept Selected" and "Reject Selected" options by default
+    $('#sodar-ss-accept-selected').hide();
```
mikkonie: If no requests are selected, these link items should be disabled, not hidden. Hiding and showing interactive UI elements depending on actions elsewhere can get quite confusing for the user. When feasible, it's better to keep those elements in fixed positions and enable/disable.

samplesheets/templates/samplesheets/irods_request_accept_form.html
```diff
@@ -17,50 +17,70 @@ <h3>Accept iRODS Delete Request</h3>
...
-      <div class="mt-3 mb-0">
-        The collection to be deleted contains the following data objects:
-        <ul class="mb-0">
-          {% for o in affected_objects %}
-            <li><code class="text-danger">{{ o.path|trim_base_path:irods_request.path }}</code></li>
-          {% endfor %}
-        </ul>
+  {% if irods_request_data %}
+    <div class="alert alert-danger">
+      <strong>Warning:</strong> Accepting the request will delete the associated
```
mikkonie: Since we can accept multiple iRODS requests now, this should be changed into "Accepting a request.."

samplesheets/templates/samplesheets/irods_request_accept_form.html
```diff
@@ -17,50 +17,70 @@ <h3>Accept iRODS Delete Request</h3>
 </div>
```
mikkonie, line 21: If accepting multiple requests, the HTML title and the subtitle in `sodar-subtitle-container` should say `Accept {number_of_requests} iRODS Delete Requests`.

samplesheets/templates/samplesheets/irods_request_accept_form.html
```diff
@@ -17,50 +17,70 @@ <h3>Accept iRODS Delete Request</h3>
...
-        The collection to be deleted comprises the following subcollections:
-        <ul class="mb-0">
-          {% for o in affected_collections %}
-            <li><code class="text-danger">{{ o.path|trim_base_path:irods_request.path }}/</code></li>
-          {% endfor %}
-        </ul>
+    {% endfor %}
+    <form method="post" enctype="multipart/form-data">
+      {% csrf_token %}
+      {{ form | crispy }}
```
mikkonie, line 42: The possible accepting of multiple requests should also be taken into consideration in the accepting field in the form, which is set in `forms.py`.

Hence, if accepting more than one request it should become `I accept the iRODS delete requests`.

samplesheets/templates/samplesheets/irods_request_accept_form.html
```diff
@@ -17,50 +17,70 @@ <h3>Accept iRODS Delete Request</h3>
...
+      {% if irods_request_data|length %}
+        <input type="hidden" name="irodsdatarequests" id="irodsdatarequests" value="{{ irods_request_uuids }}"/>
+      {% endif %}
+      <div class="row">
+        <div class="btn-group ml-auto">
+          <a role="button" class="btn btn-secondary"
+             href="{{ request.session.real_referer }}">
+            <i class="iconify" data-icon="mdi:arrow-left-circle"></i> Cancel
+          </a>
+          <button type="submit" class="btn btn-primary" id="sodar-ss-btn-delete-submit">
```
mikkonie: Since we are deleting files, this button should be colored with tthe `btn-danger` class instead of `btn-primary`. Haven't noticed this before..

samplesheets/templates/samplesheets/irods_request_accept_form.html
```diff
@@ -17,50 +17,70 @@ <h3>Accept iRODS Delete Request</h3>
...
-            <li><code class="text-danger">{{ o.path|trim_base_path:irods_request.path }}</code></li>
-          {% endfor %}
-        </ul>
+  {% if irods_request_data %}
+    <div class="alert alert-danger">
+      <strong>Warning:</strong> Accepting the request will delete the associated
+      {% if irods_request.is_collection %}collection{% else %}data object{% endif %} from iRODS.
+      This can <strong>not</strong> be undone without administrator assistance!
+    </div>
+    {% for irods_request in irods_request_data %}
```
mikkonie: I suggest we simplify this. Instead of repeating alerts for each request, I propose we simply do the following:

- Collect paths for data objects for all selected requests
    * We don't need to list subcollections, that's not really relevant info to display to the user
    * If a request targets a collection instead of a data object, include all the data objects within that collection recursively
    * Omit possible duplicates in data objects (it's possible for two requests to request the deletion of the same data object, see my comment in `views.py`)
- Present this as one list of full paths to each data object, sorted by the path
- Display the list within one alert.

samplesheets/templates/samplesheets/irods_request_accept_form.html
```diff
@@ -17,50 +17,70 @@ <h3>Accept iRODS Delete Request</h3>
...
+    <form method="post" enctype="multipart/form-data">
+      {% csrf_token %}
+      {{ form | crispy }}
+      {% if irods_request_data|length %}
+        <input type="hidden" name="irodsdatarequests" id="irodsdatarequests" value="{{ irods_request_uuids }}"/>
+      {% endif %}
+      <div class="row">
+        <div class="btn-group ml-auto">
+          <a role="button" class="btn btn-secondary"
+             href="{{ request.session.real_referer }}">
```
mikkonie: This seems to redirect back to the project sheets view instead of the delete request list. You can just redirect to `{% url 'samplesheets:irods_requests' project=project.sodar_uuid %}`.

samplesheets/templates/samplesheets/irods_requests.html
```diff
@@ -153,6 +160,24 @@ <h3>iRODS Delete Requests</h3>
                   {% endif %}
                 </div>
               </td>
+              {% if irods_webdav_enabled %}
+                  <td class="text-right" id="sodar-ss-davrods-link">
```
mikkonie: HTML `id` attributes should be unique, here we are creating multiple elements with the same ID. I think you wanted to place this `class`.

samplesheets/views.py
```diff
@@ -2329,122 +2531,120 @@ def get(self, request, *args, **kwargs):
...
-            )
-            tl_event.add_object(
-                obj=obj, label='irods_request', name=obj.get_display_name()
-            )
 
-        flow_name = 'data_delete'
-        flow_data = {'paths': [obj.path]}
-        if obj.is_data_object():
-            flow_data['paths'].append(obj.path + '.md5')
+class IrodsRequestAcceptBatchView(
```
mikkonie, line 2660: Major issue: Currently, if I try to accept a delete request targeting a collection instead of a data object, I get a crash with `VariableDoesNotExist`. This needs to be fixed of course.

We should not allow batch deletion for collection requests (see my comment in the views), but accepting a single collection request which targets a collection path should still work.

samplesheets/tests/test_views_taskflow.py
```diff
@@ -1399,45 +1399,6 @@ def test_accept_no_request(self):
             )
         self.assertEqual(response.status_code, 404)
 
-    def test_accept_invalid_form_data(self):
```
mikkonie, line 1575: You probably don't want to remove this test, maybe update it to match the current implementation.

samplesheets/views.py
```diff
@@ -885,6 +885,201 @@ def handle_alerts_deactivate(cls, irods_request, app_alerts=None):
                 )
             )
 
+    # API Helpers --------------------------------------------------------------
+
+    @classmethod
+    def process_single_accept_request(
```
mikkonie: I just realized that if we enable batch accepting for requests which target entire collections instead of data objects, this can get very messy very fast. Because what if different requests target multiple collections in the same part of the collection tree, objects within collections which are also required to be deleted, etc.

In most cases, this feature is useful for deleting several requests for individual files at once.

Hence I propose we simplify this feature as follows:

- When building the form, check for each request whether a request points to a collection
    * Use irodsbackend and `irods.collections.exists()` for this
- If request targets a collection instead of data object, disable the batch checkbox for that request

This way we shouldn't have to worry about issues regarding e.g. attemps to delete the same subcollections or data objects twice in the same batch, having to rollback in taskflow in these complex cases, etc.

samplesheets/tests/test_views_ajax_taskflow.py
```diff
@@ -531,7 +531,7 @@ def test_get_empty_coll(self):
                 data={'path': self.assay_path},
             )
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(len(response.data['irods_data']), 0)
+        self.assertEqual(len(response.data['irods_data']), 1)
```
gromdimon: Have to notice, that I do not fully understand why this values changed. The logic behind is calling 
```
with irods_backend.get_session() as irods:
    ret_data = irods_backend.get_objects(irods, self.path)
```
What responses with the data_objects. But I didn't chage anything from `irods_backend` (

samplesheets/templates/samplesheets/irods_request_accept_form.html
```diff
@@ -13,54 +13,62 @@
 
 
 <div class="container-fluid sodar-subtitle-container">
+  {% if irods_request_data|length == 1 %}
   <h3>Accept iRODS Delete Request</h3>
```
mikkonie: Make sure to include indentation in these condition blocks.

samplesheets/forms.py
```diff
@@ -450,10 +450,18 @@ class IrodsRequestAcceptForm(forms.Form):
...
-        label='I accept the iRODS delete request',
         required=True,
     )
 
+    def __init__(self, *args, **kwargs):
+        num_requests = kwargs.pop('num_requests', None)
+        super().__init__(*args, **kwargs)
+
+        if num_requests > 1:
+            self.fields['confirm'].label = 'I accept the iRODS delete requests'
```
mikkonie: For brevity, you can add the `s` character instead of repeating the entire string :)

samplesheets/tests/test_views_ajax_taskflow.py
mikkonie: The length should indeed be 0. Can you print out the return data and paste it here in the commands? I assume it's still showing that one file which should've been deleted.

I don't believe I've changed anything in irodsbackend related to this, at least on purpose. Also, all other tests work as intended. We have upgraded to a new python-irodsclient version since you started, perhaps something has changed. Maybe it has something to do with using `unlink()`..

samplesheets/views.py
```diff
@@ -2632,6 +2853,40 @@ def get_queryset(self):
...
+            )
+            return redirect(
+                reverse(
+                    'samplesheets:project_sheets',
+                    kwargs={'project': self.get_project().sodar_uuid},
+                )
+            )
+
+        if settings.IRODS_WEBDAV_ENABLED:
+            for item in response.context_data['object_list']:
```
mikkonie: This should happen in `get_context_data()`. Also, we are needlessly opening and closing iRODS connections for each data object in `object_list`. The entire loop should go inside `with irods.get_session() as session`.

samplesheets/views.py
```diff
@@ -2382,122 +2587,133 @@ def get(self, request, *args, **kwargs):
...
-                raise ex
-            messages.error(
-                self.request,
-                'Accepting iRODS data request "{}" failed: {}'.format(
-                    obj.get_display_name(), ex
-                ),
-            )
-            return redirect(redirect_url)
+    def get_form_kwargs(self):
+        "Override to pass number of requests to form"
```
mikkonie: Triple quotes missing from docstring.

samplesheets/templates/samplesheets/_list_buttons.html
```diff
@@ -0,0 +1,30 @@
...
+    <a href="{% url 'samplesheets:irods_request_create' project=project.sodar_uuid %}"
+         class="dropdown-item" role="button">
+       <i class="iconify" data-icon="mdi:plus-thick"></i> Create Request
+    </a>
+
+    {# Accept Selected #}
+    <a class="dropdown-item" id="sodar-ss-accept-selected"
+       href="javascript:{}"
+       onclick="sendRequest('{% url 'samplesheets:irods_request_accept_batch' project=project.sodar_uuid %}');">
+      <i class="iconify" data-icon="mdi:arrow-right-thick"></i> Accept Selected
```
mikkonie: The icon should be `mdi:check-bold` for consistency.

samplesheets/tests/test_views_taskflow.py
```diff
@@ -1830,6 +1839,336 @@ def test_accept_lock_failure(self):
...
+
+        with self.login(self.user_contrib):
+            self.client.post(
+                reverse(
+                    'samplesheets:irods_request_create',
+                    kwargs={'project': self.project.sodar_uuid},
+                ),
+                self.post_data,
+            )
+            self.assertEqual(IrodsDataRequest.objects.count(), 1)
```
mikkonie: Remember, we can unindent lines after `with self.login(...)` unless we need to do other stuff which requires that login. Or if we want to assert redirecting.

samplesheets/templates/samplesheets/irods_request_accept_form.html
```diff
@@ -13,54 +13,62 @@
...
+        {% for path in affected_object_paths %}
+          <li><code class="text-danger">{{ path }}</code></li>
+        {% endfor %}
+      </ul>
+    </div>
+    <form method="post" enctype="multipart/form-data">
+      {% csrf_token %}
+      {{ form | crispy }}
+      {% if irods_request_data|length %}
+        <input type="hidden" name="irodsdatarequests" id="irodsdatarequests" value="{{ irods_request_uuids }}"/>
```
mikkonie: I would rename this field as e.g. `irods_requests` to keep up with conventions.

mikkonie (changes requested):

samplesheets/tests/test_views_ajax_taskflow.py
gromdimon: I also do not see any other causes of that fail ( 
Here is an output:
Data:
 {'irods_data': [{'name': 'test2', 'type': 'obj', 'path': '/sodarZone/projects/c7/c7ddf20e-0e36-4a00-8f9c-b5afb869f349/sample_data/study_7a0621aa-f8a8-4c0a-99bc-0bb154dbf165/assay_2ffd215a-390c-4e35-9431-f3641d5cd421/test2', 'size': 0, 'modify_time': '2023-06-20 10:31', 'irods_request_status': None, 'irods_request_user': None}]}

samplesheets/views.py
```diff
@@ -2632,6 +2863,34 @@ def get_queryset(self):
         # For regular users, dispaly their own requests regardless of status
         return queryset.filter(user=self.request.user)
 
+    def build_webdav_url(self, item):
+        return f"{settings.IRODS_WEBDAV_URL}/{item.path}"
```
mikkonie: Double quotes spotted :) F-strings aren't exactly SODAR convention due to historical reasons, but eventually we should migrate SODAR Core and SODAR to them where applicable, so that part is fine!

samplesheets/views.py
```diff
@@ -2632,6 +2863,34 @@ def get_queryset(self):
...
+                request,
+                'iRODS backend not enabled',
+            )
+            return redirect(
+                reverse(
+                    'samplesheets:project_sheets',
+                    kwargs={'project': self.get_project().sodar_uuid},
+                )
+            )
+
```
mikkonie: Please remove the blank line.

samplesheets/views.py
```diff
@@ -961,6 +962,205 @@ def handle_alerts_deactivate(cls, irods_request, app_alerts=None):
                 )
             )
 
+    # API Helpers --------------------------------------------------------------
+
+    @classmethod
+    def process_single_accept_request(
```
mikkonie: To simplify and keep with conventions, we can simply name this `accept_request()`.

samplesheets/views.py
```diff
@@ -961,6 +962,205 @@ def handle_alerts_deactivate(cls, irods_request, app_alerts=None):
...
+            self.handle_alerts_deactivate(obj, app_alerts)
+
+        return {
+            'message': 'iRODS data request "{}" accepted.'.format(
+                obj.get_display_name()
+            )
+        }
+
+    @classmethod
+    def process_single_reject_request(
```
mikkonie: Similarly, this can be simply called `reject_request()`.

samplesheets/views.py
```diff
@@ -2337,37 +2537,42 @@ class IrodsRequestAcceptView(
...
 ):
-    """View for accepting an iRODS data request"""
+    """View for accepting iRODS data requests"""
 
     permission_required = 'samplesheets.manage_sheet'
     template_name = 'samplesheets/irods_request_accept_form.html'
     form_class = IrodsRequestAcceptForm
 
+    def get_form_kwargs(self):
+        "Override to pass number of requests to form"
```
mikkonie: Triple quotes missing.

mikkonie (changes requested):
Good work, I just had some minor cleanup/style comments. Now I will look into that failing test case..

docs_manual/source/app_samplesheets_irods_delete.rst
```diff
@@ -99,3 +100,15 @@ user will be informed of rejection.
     Accepting delete requests will delete the associated file(s) from iRODS with
     no possibility for undoing the action! Each request should be reviewed
     carefully.
+
+
+Multiple Requests Accepting and Rejecting
```
mikkonie: Please rename into "Accepting and Rejecting Multiple Requests".

docs_manual/source/app_samplesheets_irods_delete.rst
```diff
@@ -99,3 +100,15 @@ user will be informed of rejection.
...
+=========================================
+
+In addition to accepting or rejecting requests one by one, you can also accept
+or reject multiple requests at once. This is done by selecting the requests you
+want to accept or reject by clicking the checkboxes on the leftmost column of
+the request list. Once you have selected the requests, click the
```
mikkonie, line 111: Please add a note that requests for entire collections must be accepted or rejected individually.

samplesheets/forms.py
```diff
@@ -458,10 +458,17 @@ class IrodsRequestAcceptForm(forms.Form):
     """Form accepting an iRODS delete request."""
 
     confirm = forms.BooleanField(
-        label='I accept the iRODS delete request',
         required=True,
```
mikkonie: I think we have extra line breaks here after removing the label :)

samplesheets/templates/samplesheets/irods_requests.html
```diff
@@ -58,6 +58,10 @@
         display: none;
       }
     }
+
+    .sodar-ss-davrods-link {
```
mikkonie: I don't think this is needed, as the link is supposed to be pulled to the right end of the column. Didn't we already remove this once? Some merge problem maybe?

samplesheets/templates/samplesheets/irods_request_accept_form.html
```diff
@@ -13,54 +13,62 @@
...
-  <form method="post" enctype="multipart/form-data">
-    {% csrf_token %}
-    {{ form | crispy }}
+    </div>
     <div class="row">
-      <div class="btn-group ml-auto">
+      <div class="ml-auto btn-group">
         <a role="button" class="btn btn-secondary"
-           href="{{ request.session.real_referer }}">
+           href="{% url 'samplesheets:irods_requests' project=project.sodar_uuid %}">
```
mikkonie: It seems we unnecessarily repeat the cancel button HTML in the template. We can simply hide the accept button and present the proper alert before it if `irods_request_data` is not available.

samplesheets/templates/samplesheets/irods_requests.html
```diff
@@ -91,7 +92,16 @@ <h3>iRODS Delete Requests</h3>
             <th>User</th>
             <th>Created</th>
             <th>Status</th>
-            <th></th>
+            {% if irods_webdav_enabled %}
+              <th></th>
+            {% endif %}
+            <th id="sodar-ss-list-header-select">
```
mikkonie: I would make the ID a bit more specific, e.g. `sodar-ss-request-header-select`.

samplesheets/templates/samplesheets/irods_requests.html
```diff
@@ -91,7 +92,16 @@ <h3>iRODS Delete Requests</h3>
...
+              {% if can_manage_request %}
+                <input class="sodar-checkbox" type="checkbox"
+                       onchange="checkAll(this)" name="check_all"
+                       id="sodar-ss-check-all" title="Check/uncheck all" />
```
mikkonie: Same here, I would define the ID as e.g. `sodar-ss-request-check-all`.

samplesheets/tests/test_views_taskflow.py
```diff
@@ -1090,9 +1091,14 @@ def setUp(self):
         self.assay_path = self.irods_backend.get_path(self.assay)
         self.path = os.path.join(self.assay_path, TEST_FILE_NAME)
         self.path_md5 = os.path.join(self.assay_path, f'{TEST_FILE_NAME}.md5')
+        # Second file
+        self.path2 = os.path.join(self.assay_path, TEST_FILE_NAME2)
+        self.path2_md5 = os.path.join(self.assay_path, f'{TEST_FILE_NAME2}.md5')
         # Create objects
         self.file_obj = self.irods.data_objects.create(self.path)
         self.md5_obj = self.irods.data_objects.create(self.path_md5)
+        self.file_obj2 = self.irods.data_objects.create(self.path2)
```
mikkonie: Are `self.file_obj2` and `self.md5_obj2` actually used anywhere?

samplesheets/tests/test_views_taskflow.py
```diff
@@ -1830,6 +1838,330 @@ def test_accept_lock_failure(self):
...
+                },
+            )
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(
+            response.context['irods_request_data'][0],
+            IrodsDataRequest.objects.first(),
+        )
+        self.assertEqual(
+            response.context['affected_object_paths'][0],
+            coll_path,
```
mikkonie: Looks like an unnecessary line break here :)

samplesheets/views.py
```diff
@@ -2538,37 +2561,42 @@ class IrodsRequestAcceptView(
...
+        kwargs = super().get_form_kwargs()
+        kwargs['num_requests'] = 1
+        return kwargs
+
     def get_context_data(self, *args, **kwargs):
         context_data = super().get_context_data(*args, **kwargs)
         obj = IrodsDataRequest.objects.filter(
             sodar_uuid=self.kwargs['irodsdatarequest']
         ).first()
+        context_data['irods_request_data'] = []
```
mikkonie: `irods_request_data` seems vague. Could this be `irods_requests`, or maybe `request_objects`?

samplesheets/views.py
```diff
@@ -2583,59 +2611,162 @@ def get(self, request, *args, **kwargs):
...
+        if form.is_valid():
+            timeline = get_backend_api('timeline_backend')
+            taskflow = get_backend_api('taskflow')
+            app_alerts = get_backend_api('appalerts_backend')
+            project = self.get_project()
+            batch = self.get_irods_request_objects()
+            if not batch:
+                messages.error(
+                    self.request,
+                    IRODS_NO_REQ_MSG,
```
mikkonie: Unnecessary line breaks :)

samplesheets/views.py
```diff
@@ -2741,6 +2939,33 @@ def get_queryset(self):
...
+            item.webdav_url = None
+        # Check if the item is a collection
+        item.is_collection = irods_session.collections.exists(item.path)
+
+    def get(self, request, *args, **kwargs):
+        irods_backend = get_backend_api('omics_irods')
+        if not irods_backend:
+            messages.error(
+                request,
+                'iRODS backend not enabled',
```
mikkonie: And some more unneeded line breaking :)

samplesheets/views.py
```diff
@@ -2538,37 +2561,42 @@ class IrodsRequestAcceptView(
     IrodsRequestModifyMixin,
     FormView,
 ):
-    """View for accepting an iRODS data request"""
+    """View for accepting iRODS data requests"""
```
mikkonie, line 2564: One major bug: when I click on "Accept Request" on a request where the path is a collection, I get the following error: `'dict' object has no attribute 'path'`

Please fix and also introcude test(s) which ensure the form renders and works properly on requests which target collections instead of data objects.

mikkonie (changes requested):
Reviewed, one rather serious bug + mostly minor cleanup things.

samplesheets/templates/samplesheets/irods_requests.html
gromdimon: I thought, that there are 2 buttons, that are alligned to the right side. And if I remove ident, then they look like that:
image
So, should I remove that class?

samplesheets/tests/test_views_taskflow.py
gromdimon: Yep, with help of them the `self.path2` and `self.path2_md5` are being formed (and later in testing Batch views being used)

samplesheets/templates/samplesheets/irods_requests.html
mikkonie: Ah, good point. Yes, remove it from the CSS, but add a `ml-1` class to the rightmost button. Bootstrap has these nice margin classes, let's use them where ever applicable.

samplesheets/templates/samplesheets/irods_request_accept_form.html
```diff
@@ -13,54 +13,56 @@
...
+  {% if request_objects %}
+    <div class="alert alert-danger">
+      <strong>Warning:</strong> Accepting {% if request_objects|length == 1 %}a request{% else %}requests{% endif %}
+      will delete the associated data objects from iRODS.
+      This can <strong>not</strong> be undone without administrator assistance!
+    </div>
+    <div class="alert alert-danger">
+      <strong>Info:</strong> The following {% if affected_object_paths|length > 1 %}
+      {{ affected_object_paths|length }} data objects
+      {% else %}data object{% endif %} will be deleted:
```
mikkonie, line 33: I noticed a small bug: collections are also listed as "data objects" here. In the screenshot below, I'm accepting the deletion of the collection `delete-test` which contains 2 data objects. The template lists "3 data objects", one of which is the collection itself. We should only list data objects here, as deleting collections themselves doesn't really matter.

2023-06_delete_request_bug

samplesheets/tests/test_views_taskflow.py
```diff
@@ -1829,6 +1837,385 @@ def test_accept_lock_failure(self):
...
+                ),
+            )
+
+        obj.refresh_from_db()
+        self.assertEqual(obj.status, 'ACCEPTED')
+        self._assert_alert_count(CREATE_ALERT, self.user, 0)
+        self._assert_alert_count(CREATE_ALERT, self.user_delegate, 0)
+        self._assert_alert_count(ACCEPT_ALERT, self.user, 0)
+        self._assert_alert_count(ACCEPT_ALERT, self.user_delegate, 0)
+        self.assert_irods_obj(coll_path, False)
```
mikkonie: Note that this is supposed to be used for data objects only, not collections. I believe this will always return False, as not data object with the path of a collection will exist. Use `self.irods.collections.exists()` instead.

Also make sure this has been done elsewhere in tests when asserting the existence of a collection.

samplesheets/tests/test_views_taskflow.py
```diff
@@ -1829,6 +1837,385 @@ def test_accept_lock_failure(self):
...
+                            for irods_request in IrodsDataRequest.objects.all()
+                        ]
+                    ),
+                },
+            )
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(
+            response.context['affected_object_paths'],
+            sorted(
+                set(
```
mikkonie, line 1893: Is `set()` actually needed here? Are there multiple requests with the same path?

samplesheets/tests/test_views_taskflow.py
```diff
@@ -1829,6 +1837,385 @@ def test_accept_lock_failure(self):
...
+class TestIrodsRequestAcceptBatchView(TestIrodsRequestViewsBase):
+    """Tests for IrodsRequestAcceptBatchView"""
+
+    def test_render(self):
+        """Test rendering IrodsRequestAcceptBatchView"""
+        self.assert_irods_obj(self.path)
+        self.assert_irods_obj(self.path_md5)
+
+        with self.login(self.user_contrib):
+            self.client.post(
```
mikkonie: Since we already test the creation in its dedicated test class, you could simplify accept/reject/delete tests by replacing these creation post requests by simply creating the request object in the SODAR database. For this you can use `IrodsDataRequestMixin.make_irods_data_request()`.

samplesheets/tests/test_views_taskflow.py
```diff
@@ -1829,6 +1837,385 @@ def test_accept_lock_failure(self):
...
+                    'samplesheets:irods_request_create',
+                    kwargs={'project': self.project.sodar_uuid},
+                ),
+                self.post_data2,
+            )
+            self.assertEqual(IrodsDataRequest.objects.count(), 2)
+
+        with self.login(self.user):
+            response = self.client.post(
+                reverse(
```
mikkonie: The tests could be condensed by setting this repeated URL as e.g. `self.url`.

mikkonie (changes requested):
Looking pretty good, found one small bug and had suggestions for test improvements and other minor things.

samplesheets/tests/test_views_taskflow.py
gromdimon: When I created multiple requests refering to the same path directly from shell, there was possibility of accepting multiple requests with only one path (so it should be filtered with set). I thought, that different users might want to delete the same object and so would create redundant requests, which in that case should be filtered. 

If it's not the case I can delete set() :)

samplesheets/tests/test_views_taskflow.py
gromdimon: It is way better approach, but after changing the current implementation, most alert checks start failing. It won't be a problem if that would be create alerts after creation, but there are use-cases of checking create alerts after deleting, accepting and rejecting requests (it also shows different behaviour if there are multiple requests).

I see couple of possibilities to fix that:
- Avoid alert tests and just replace creations.
- Avoid direct creation method and let it be as it is.
- Write new tests for alert testing with old implementation and change every other test to new one.

What do you think on that?

samplesheets/tests/test_views_taskflow.py
mikkonie: Ah, good point. You can leave the implementation as is.

samplesheets/tests/test_views_taskflow.py
mikkonie: It should be enough if we test the create alerts in the create view tests. They should not be affected by delete/accept/reject requests, right? The create alerts shouldn't be e.g. modified afterwards by any further actions, if I'm correct?

There are indeed things which affect how create alerts are generated, but these should be tested in `TestIrodsRequestCreateView`. There we should, of course, use a `POST` request to create the request AND the necessary alerts, while in the subsequent test classes the creation can be simplified and we only look for the alerts specific to that view under testing.

This will also make the tests more atomic and consistent, as we specifically test the functionality of the view in question.

So my suggestion would be:

- Test all necessary create alert actions in `TestIrodsRequestCreateView`
- In test classes for accept/reject/delete views, simplify the request creation and only test for the alerts relevant to the view in question

It's possible I'm missing some case where this isn't feasible so if you disagree, please let me know.

samplesheets/templates/samplesheets/irods_request_accept_form.html
mikkonie: This works fine now, good job fixing it!

samplesheets/tests/test_views_taskflow.py
```diff
@@ -1090,9 +1091,14 @@ def setUp(self):
         self.assay_path = self.irods_backend.get_path(self.assay)
         self.path = os.path.join(self.assay_path, TEST_FILE_NAME)
```
mikkonie: One further minor point: I would recommend renaming this and related paths to e.g. `self.obj_path*`. It specifies this is a path for a data object, not a collection. It makes the tests more readable.

mikkonie (changes requested):
This is looking very good. One of your questions was a fair point and that is considered resolved. I answered the other one, please let me know if you disagree with my response.

I also had one very minor test refactoring request, other than these two issues we seem to be good to merge!

mikkonie (approved):
Looking good, no notes!

PR #1696, opened 2023-06-06
`"nothing to do"` check for landingzones validation/moving

### LINKED ISSUE #339: Add "nothing to do" check for landingzones validation/moving
At the moment, it's possible to "successfully" validate of move a landing zone with 0 files in it. While this is technically a successful operation since nothing can go wrong, it doesn't make much sense :)

Instead, we should check for the amount of files in the zone and return to the zone list with a "nothing to do" message.

An alternative would be to disable the "validate" / "validate and move" menu items until the iRODS stats badge displays files, but this might be inpractical due to the user having to wait for the update before proceeding, so I don't like that.

## Spec

- Check for zone file count in `ZoneMoveView.get()`
    * Hint: Use `IrodsAPI.get_object_stats()` witth `IrodsAPI.get_path(zone)`
- IF `file_count == 0`
    * THEN  add Django message with `info` level and message "No files in landing zone, nothing to do."
    * AND redirect user back to project zone list
- Add test(s) to `landingzones.tests.test_views_taskflow`

gromdimon:
### TODO:
- [x] Implement 
- [x] Test


landingzones/tests/test_permissions.py
```diff
@@ -27,10 +33,15 @@
 
 # Local constants
 SHEET_PATH = SHEET_DIR + 'i_small.zip'
+TEST_OBJ_NAME = 'test1.txt'
 
 
 class TestLandingZonePermissionsBase(
-    LandingZoneMixin, SampleSheetIOMixin, TestProjectPermissionBase
+    LandingZoneMixin,
```
gromdimon, line 34: After implementation of `nothing to do` check the whole class was failing in CI. The reason was the lack of files (what caused `nothing to do` case). So I found a possible solution: add files to the landingzones. However it got very messy because of tons of dependencies from taskflow( Maybe there could be another implementation?

landingzones/tests/test_permissions.py
```diff
@@ -15,7 +17,9 @@
     ZONE_TITLE,
     ZONE_DESC,
 )
-
+from taskflowbackend.tests.base import (
+    TaskflowAPIPermissionTestBase,
```
mikkonie: This import could be on one line.

landingzones/tests/test_permissions.py
```diff
@@ -199,313 +206,3 @@ def test_zone_create_disable(self):
         ]
         self.assert_response(url, good_users, 200)
         self.assert_response(url, bad_users, 302)
-
-    def test_zone_update(self):
```
mikkonie, line 203: Unless I'm mistaken, it should be possible to keep the zone update and delete permission tsts in `test_permissions`, as no taskflow creation is needed to access these views.

landingzones/tests/test_permissions_taskflow.py
```diff
@@ -0,0 +1,388 @@
+"""Tests for UI view permissions with taskflow"""
+
+from django.test import override_settings
+from django.urls import reverse
+
+# Projectroles dependency
+from projectroles.models import SODAR_CONSTANTS
+from projectroles.tests.test_permissions import (
+    TestPermissionMixin,
```
mikkonie: This import could be on one line.

landingzones/tests/test_permissions_taskflow.py
```diff
@@ -0,0 +1,388 @@
...
+)
+
+from landingzones.tests.test_views_taskflow import LandingZoneTaskflowMixin
```
mikkonie: Make sure the imports are in the expected order:

- projectroles
- samplesheets
- taskflowbackend
- landingzones

landingzones/tests/test_views.py
```diff
@@ -267,7 +267,7 @@ def test_render(self):
                     kwargs={'landingzone': self.landing_zone.sodar_uuid},
                 )
             )
-        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.status_code, 302)
```
mikkonie: You can remove this test from `test_views` and instead add an extra test in `test_views_taskflow` to ensure the behaviour with a data object in the landing zone.

landingzones/views.py
```diff
@@ -715,6 +716,31 @@ def get_context_data(self, *args, **kwargs):
...
         zone = LandingZone.objects.get(sodar_uuid=self.kwargs['landingzone'])
+        try:
+            irods_backend = get_backend_api('omics_irods')
+            path = irods_backend.get_path(zone)
+            with irods_backend.get_session() as irods:
+                stats = irods_backend.get_object_stats(irods, path)
+                if stats['file_count'] == 0:
+                    messages.info(
+                        request,
+                        ZONE_MOVE_NO_FILES,
```
mikkonie: It seems there are some extra line breaks here.

mikkonie (changes requested):
Looking good, had some minor comments.

landingzones/tests/test_permissions.py
```diff
@@ -27,10 +30,14 @@
...
 SHEET_PATH = SHEET_DIR + 'i_small.zip'
+TEST_OBJ_NAME = 'test1.txt'
 
 
 class TestLandingZonePermissionsBase(
-    LandingZoneMixin, SampleSheetIOMixin, TestProjectPermissionBase
+    LandingZoneMixin,
+    SampleSheetIOMixin,
+    TestPermissionMixin,
+    TaskflowAPIPermissionTestBase,
```
mikkonie: Since these are the non-taskflow permission tests, this should be based on `TestProjectPermissionBase`. That base view already uses `TestPermissionMixin`, so re-including that is not necessary.

landingzones/tests/test_permissions_taskflow.py
```diff
@@ -0,0 +1,222 @@
...
+TEST_OBJ_NAME = 'test1.txt'
+
+
+class TestLandingZonePermissionTaskflowBase(
+    LandingZoneMixin,
+    LandingZoneTaskflowMixin,
+    SampleSheetIOMixin,
+    TestPermissionMixin,
+    TaskflowTestMixin,
+    TaskflowAPIPermissionTestBase,
```
mikkonie, line 44: This should be based on an UI taskflow permission test base class, but it looks like one doesn't currently exist. I'll add a separate ticket for this. For now, if this works, feel free to keep it as is.

mikkonie (changes requested):

mikkonie (approved):
Looks good, no notes!

PR #1701, opened 2023-06-12
fix handling of empty strings in ISA-tab templates

### LINKED ISSUE #1700: Empty string in sheet template field choices displays empty options dialogue
When an ISA-tab template includes a choice variable with one of the items being an empty string, the drop-down menu in SODAR's Samplesheets > Create from template for that variable is empty i. e. displays no items at all. This can be seen with the current version of the MS Metabolomics Biocrates kit ISA-tab template. Variables:

- Biocrates metidq version
- Metaquac version

Expected behaviour: Drop-down menu displays empty line for empty string choices.

sellth:
Handling of empty strings in cookiecutter templates currently overwrites the entire kwargs object when probably only `required=False` should be added.

fixes #1700 

Interesting side note: When this bug was triggered, `django.forms.*Field` was called without a label argument. The resulting view showed the respective label anyway, but with a capital first letter and spaces instead of underscores. Not sure how this happens, but it's actually a nicer looking label.

### DIFF
--- a/samplesheets/forms.py
+++ b/samplesheets/forms.py
@@ -192,13 +192,13 @@ class SheetTemplateCreateForm(forms.Form):
                 self.initial[k] = clean_sheet_dir_name(project.title)
             elif isinstance(v, str):
                 if not v:  # Allow empty value if default is not set
-                    field_kwargs = {'required': False}
+                    field_kwargs['required'] = False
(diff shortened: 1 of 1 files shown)


mikkonie:
Well spotted, that's a quite obvious and terrible mistake. My bad :)

PR #1702, opened 2023-06-12
Path clipboard copy button for landing zone irods collection list

### LINKED ISSUE #1282: Add path clipboard copy buttons to collections in landing zone iRODS collection list
This seems like an obvious feature now that we also display collections in the list.

## Spec

- In `irodsbackend.js` under "Link list popup"...
- IF `obj['type'] === 'coll'`
- THEN add button into first column to copy path to clipboard
    * Pull button right so it aligns with the right side of the column
    * Use the standard clipboard copying button style from the zone list
        + `btn btn-secondary sodar-list-btn`, icon as `mdi:console-line`, title as "Copy iRODS path into clipboard"
    * Use `obj['path']` as the value to be copied
- I think we can skip UI tests for this, as testing the clipboard is painful and this is a very simple feature
- **Bonus:** If you want to clean up and refactor the JQuery into something more pretty, please do! This is old and quite ugly :)

gromdimon:
### TODO:
- [x] Implement feature
- [ ] Refactor code


landingzones/templates/landingzones/project_zones.html
```diff
@@ -21,6 +21,10 @@
         table-layout: fixed;
     }
 
+    .sodar-lz-copy-btn {
+      margin-left: 5px;
```
mikkonie: Since we use Bootstrap, you should be able to also do this by adding e.g. `ml-1` or `ml-2` to the classes of the button :) But since we want to pin this to the right end of the column, this is not needed anyway.

irodsbackend/static/irodsbackend/js/irodsbackend.js
```diff
@@ -254,11 +254,18 @@ $(document).ready(function() {
                         colSpan = '4';
                         icon = 'mdi:folder-open';
                         toolTip = 'Collection';
+
+                        var copyButton = $('<button>')
```
mikkonie: The buttons are not currently pulled to the right side of the table column as specified in the spec. Instead, they are directly behind the collection or file title. If `pull-right` doesn't work, maybe `ml-auto` does. I forget which but you can find examples from existing tables and experiment.

irodsbackend/static/irodsbackend/js/irodsbackend.js
```diff
@@ -254,11 +254,18 @@ $(document).ready(function() {
...
+                            .addClass('btn btn-secondary sodar-list-btn sodar-irods-copy-btn sodar-lz-copy-btn')
```
mikkonie: It seems the bootstrap tooltips and the "copied!" notification from sodar-irods-copy-btn are not working. The `data-tooltip="tooltip" data-placement="top"` attributes seem to be missing.

mikkonie:
The current commit seems to have a fatal bug. If i try to open a modal for a landing zone which has data objects in it, the modal fails to open and javascript console says:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'prop')
    at Object.<anonymous> (irodsbackend.js:294:36)
    at Function.each (jquery-3.5.1.min.js:2:2976)
    at Object.<anonymous> (irodsbackend.js:263:19)
    at c (jquery-3.5.1.min.js:2:28294)
    at Object.fireWith [as resolveWith] (jquery-3.5.1.min.js:2:29039)
    at l (jquery-3.5.1.min.js:2:79800)
    at XMLHttpRequest.<anonymous> (jquery-3.5.1.min.js:2:82254)
```

Landing zones with only collections work OK. At least that's the only difference I can see right now.

Hot tip: I always keep some landing zones with files in my dev site to quickly test things like this :)

mikkonie:
There's are two issues with the current version.

1. The crash with a landing zone containing data object still occurs. Please let me know if you have trouble reproducing it and I'll try to help.
2. The clipboard copy button now seems to be right aligned to the entire length of the table. It should be on the right side of the 1st column of the table. See screenshot below. There seems to be a `colspan` attribute if the table row contains a collection. We can change that implementation to have all the columns intact, or, have the colspan affect columns from the 2nd one onwards.

2023-06_zone_path_copy_bug

irodsbackend/static/irodsbackend/js/irodsbackend.js
```diff
@@ -254,11 +254,18 @@ $(document).ready(function() {
...
+                            .append($('<i>').addClass('iconify').attr('data-icon', 'mdi:console-line'))
```
mikkonie: One thing I forgot to mentione before: here we should actually use `mdi:clipboard-text-multiple`. It is generally used as the "copy to clipboard" symbol in SODAR. In the case of multiple copy links we differentiate between the console etc, but here that's not needed.

mikkonie (changes requested):
I was dumb and left my previous review issues as regular comments in the PR instead of an actual review, so you may have missed them. Please see the previous comments when updating this :)

irodsbackend/static/irodsbackend/js/irodsbackend.js
```diff
@@ -247,18 +270,24 @@ $(document).ready(function() {
...
+                            '<i class="iconify" data-icon="mdi:clipboard-text-multiple"></i>' +
+                            '</button>';
                     }
-                    var iconHtml = '<i class="iconify mr-1" data-icon="'+ icon +'"' +
-                        'title="'+ toolTip +'"></i>';
 
-                    body += '<tr><td colspan="' + colSpan + '">' + iconHtml + objLink + '</td>';
+                    var iconHtml = '<i class="iconify mr-1" data-icon="' + icon + '"' +
+                        ' title="' + toolTip + '"></i>';
+                    body += '<tr><td colspan="' + colSpan + '">' + iconHtml + objLink + copyButton + '</td>';
```
mikkonie: There are three issues with the current functionality:

- The copy button is missing for data objects
- Cells for collections are not drawn beyond the first column
- The headers no longer correspond to table rows (probably a colspan issue?)

2023-06_zone_path_copy_bug2

irodsbackend/static/irodsbackend/js/irodsbackend.js
```diff
@@ -247,18 +270,24 @@ $(document).ready(function() {
...
+
+                    var copyButton = '';
                     if (obj['type'] === 'coll') {
-                        colSpan = '4';
-                        icon = 'mdi:folder-open';
-                        toolTip = 'Collection';
+                        copyButton = '<button class="btn btn-secondary sodar-list-btn pull-right" ' +
+                            'id="sodar-irods-copy-btn-' + obj['name'] + '" ' +
+                            'title="Copy iRODS path into clipboard" data-tooltip="tooltip" ' +
+                            'data-placement="top" onclick="copy_path(\'' + obj['path'] + '\', \'sodar-irods-copy-btn-' + obj['name'] + '\')">' +
```
mikkonie: While we don't enforce a specific line length for JQuery files at the moment, super long lines like this one should be broken down.

mikkonie (changes requested):

mikkonie (approved):

PR #1712, opened 2023-06-15
Enable landing zone controls for superuser if zone is locked

### LINKED ISSUE #1607: Enable landing zone controls for superuser if zone is locked
A very minor issue, but this would speed up maintenance and debugging situations when we end up with a "stuck" landing zone.

## Spec

- IF user is superuser and zone is locked, THEN do not disable any landing zone buttons or the dropdown
   * Hint: See row 59 in `landingzones.js`
- Provide superuser status to JQuery via the zone list template
    * Don't worry about perm checks, perms will still be checked in the related views (and iRODS) even if someone injects a `superuser=true` status in the DOM :)
- Add UI test to ensure the buttons stay operational for superuser even if the zone is not ACTIVE or FAILED

### LINKED ISSUE #1657: Add DELETING to locked states in landingzones UI
For some reason the locked status is missing from the `DELETING` status, even though no actions are allowed during this state.

See line 40 in `landingzones.js`.

Very simple item, can be done together with #1607.

gromdimon:
### TODO:
- [x] Add superuser variable in Jquery
- [x] Implement feature
- [x] Test

### DIFF
--- a/landingzones/templates/landingzones/project_zones.html
+++ b/landingzones/templates/landingzones/project_zones.html
@@ -163,7 +163,7 @@
 
   <!-- Superuser status variable -->
   <script>
-    var isSuperuser = {{ superuser|lower }};
+    var isSuperuser = {{ request.user.is_superuser }};
   </script>
 
   <!-- Landingzones Javascript -->
--- a/landingzones/templates/landingzones/project_zones.html
+++ b/landingzones/templates/landingzones/project_zones.html
@@ -159,10 +159,7 @@
   <script type="text/javascript">
     window.statusInterval = {{ zone_status_interval }} * 1000;
     window.irodsShowChecksumCol = true;
-  </script>
-
-  <!-- Superuser status variable -->
-  <script>
+    <!-- Superuser status variable -->
     var currentUserURL = "{% url 'projectroles:ajax_user_current' %}";
   </script>
 
--- a/landingzones/templates/landingzones/project_zones.html
+++ b/landingzones/templates/landingzones/project_zones.html
@@ -163,7 +163,7 @@
 
   <!-- Superuser status variable -->
   <script>
-    var isSuperuser = {{ request.user.is_superuser }};
+    var currentUserURL = "{% url 'projectroles:ajax_user_current' %}";
   </script>
 
   <!-- Landingzones Javascript -->
--- a/landingzones/templates/landingzones/project_zones.html
+++ b/landingzones/templates/landingzones/project_zones.html
@@ -161,6 +161,11 @@
     window.irodsShowChecksumCol = true;
   </script>
 
+  <!-- Superuser status variable -->
+  <script>
+    var isSuperuser = {{ superuser|lower }};
+  </script>
+
   <!-- Landingzones Javascript -->
   <script type="text/javascript" src="{% static 'landingzones/js/landingzones.js' %}"></script>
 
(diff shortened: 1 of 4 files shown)


landingzones/templates/landingzones/project_zones.html
```diff
@@ -161,6 +161,11 @@ <h3><i class="iconify" data-icon="mdi:briefcase-upload"></i> Landing Zones</h3>
     window.irodsShowChecksumCol = true;
   </script>
 
+  <!-- Superuser status variable -->
+  <script>
+    var isSuperuser = {{ superuser|lower }};
```
mikkonie: We can simply do `{{ request.user.is_superuser }}` here, no need to set a separate variable in context.

landingzones/views.py
```diff
@@ -405,6 +405,8 @@ class ProjectZoneView(
 
     def get_context_data(self, *args, **kwargs):
         context = super().get_context_data(*args, **kwargs)
+        # # Superuser status
+        context['superuser'] = self.request.user.is_superuser
```
mikkonie: This we can omit, see my comment in the template.

landingzones/tests/test_ui.py
```diff
@@ -333,3 +333,54 @@ def test_zone_buttons_archive(self):
...
+                By.CLASS_NAME, 'sodar-lz-zone-title'
+            ).get_attribute('class'),
+        )
+        self.assertNotIn(
+            'text-muted',
+            zone.find_element(
+                By.CLASS_NAME, 'sodar-lz-zone-status-info'
+            ).get_attribute('class'),
+        )
+        self._assert_btn_enabled(
```
mikkonie: We can shorten this a bit by looping the assertion on a list of the class names.

landingzones/templates/landingzones/project_zones.html
mikkonie: As a general note, in the future we'll want to replace things like this in a future with a user status ajax call which returns this info.

For anything serious, we don't want to provide this info to javascript in the template: this can be edited in run-time and "gain" superuser access. However, in this case it's OK, because the user can not really do anything dangerous with that info.

mikkonie (changes requested):
This works and looks ok, goot job. I had a few minor comments.

mikkonie:
This works as specified, but one thing I didn't think of were the information security issues with passing info such as `is_superuser` to the client from the template.

These variables can be changed in run-time in the inspector, so technically any user could "gain" superuser access. In this particular case it's not super critical, but technically it's not something we should do.

I'd like you to try and refactor this in the following way:

- In `landingzones.js`, once the document is ready, do an ajax call to `CurrentUserRetrieveAjaxView`
    * See SODAR Core for that view
- Store the superuser status as a javascript variable from the view's response
- ..and, you guessed it, use that variable instead :)

This way the info can't be forged.

gromdimon:
I found out, that another issue could be implemented here as well
#1657

landingzones/templates/landingzones/project_zones.html
```diff
@@ -161,6 +161,11 @@ <h3><i class="iconify" data-icon="mdi:briefcase-upload"></i> Landing Zones</h3>
     window.irodsShowChecksumCol = true;
   </script>
 
+  <!-- Superuser status variable -->
+  <script>
```
mikkonie: Nice touch providing the URL for JQuery via the template, it's much better than hardcoding it.

We can include the URL in the previous `<script>` block which already provides some settings to the client side Javascript. No need to declare a new one.

Other than that, no notes for this pull request. Good work :)

mikkonie (changes requested):

PR #1713, opened 2023-06-19
Add irodsinfo REST API endpoint to retrieve iRODS configuration

### LINKED ISSUE #1685: Add irodsinfo REST API endpoint to retrieve iRODS configuration
This would enable external tools such as cubi-tk to automatically fetch a valid iRODS configuration if SODAR API access was already established. Not relying on the user manually downloading `irods_environment.json` into the right location improves UX during onboarding of new users/machines.

related: bihealth/sodar-cli/issues/6

gromdimon:
### TODO:
- [x] Add REST API endpoint to irodsinfo.views_api
- [x] Refactor config generating in IrodsConfigView into a mixin
- [x] Add permission and view tests for the API view
- [x] Add irodsinfo REST API docs to manual

### DIFF
--- a/irodsinfo/tests/test_permissions.py
+++ b/irodsinfo/tests/test_permissions.py
@@ -3,28 +3,25 @@
 from django.urls import reverse
 
 # Projectroles dependency
-from projectroles.tests.test_permissions import TestProjectPermissionBase
+from projectroles.tests.test_permissions import TestPermissionBase
 
 
-class TestIrodsinfoPermissions(TestProjectPermissionBase):
+class TestIrodsinfoPermissions(TestPermissionBase):
     """Tests for irodsinfo UI view permissions"""
+
+    def setUp(self):
+        # Create users
+        self.superuser = self.make_user('superuser')
+        self.superuser.is_superuser = True
+        self.superuser.save()
+        self.regular_user = self.make_user('regular_user')
+        # No user
+        self.anonymous = None
 
     def test_irods_info(self):
         """Test permissions for IrodsInfoView"""
         url = reverse('irodsinfo:info')
-        good_users = [
-            self.superuser,
-            self.user_owner_cat,
-            self.user_delegate_cat,
-            self.user_contributor_cat,
-            self.user_guest_cat,
-            self.user_finder_cat,
-            self.user_owner,
-            self.user_delegate,
-            self.user_contributor,
-            self.user_guest,
-            self.user_no_roles,
-        ]
+        good_users = [self.superuser, self.regular_user]
         bad_users = [self.anonymous]
         self.assert_response(url, good_users, 200)
         self.assert_response(url, bad_users, 302)
@@ -32,19 +29,7 @@ class TestIrodsinfoPermissions(TestProjectPermissionBase):
     def test_irods_config(self):
         """Test permissions for IrodsConfigView"""
         url = reverse('irodsinfo:config')
-        good_users = [
-            self.superuser,
-            self.user_owner_cat,
-            self.user_delegate_cat,
-            self.user_contributor_cat,
-            self.user_guest_cat,
-            self.user_finder_cat,
-            self.user_owner,
-            self.user_delegate,
-            self.user_contributor,
-            self.user_guest,
-            self.user_no_roles,
-        ]
+        good_users = [self.superuser, self.regular_user]
         bad_users = [self.anonymous]
         self.assert_response(url, good_users, 200)
         self.assert_response(url, bad_users, 302)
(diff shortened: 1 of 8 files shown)


irodsinfo/tests/test_permissions.py
```diff
@@ -3,33 +3,48 @@
 from django.urls import reverse
 
 # Projectroles dependency
-from projectroles.tests.test_permissions import TestPermissionBase
+from projectroles.tests.test_permissions import TestProjectPermissionBase
 
 
-class TestIrodsinfoPermissions(TestPermissionBase):
+class TestIrodsinfoPermissions(TestProjectPermissionBase):
```
mikkonie: Irodsinfo is a site app, so it doesn't have a project context. Hence these tests should be achievable with `TestPermissionBase`.

Since there is no project context, we also only need to test for:

- Regular user (non-superuser)
- Superuser
- Anonymous

In the future, we should set up a site app permission test base class to SODAR Core to simplify this. I'll add an issue for that.

Meanwhile, you can just init the required users manually in `setUp()` just like the original tests did.

irodsinfo/tests/test_permissions_api.py
```diff
@@ -0,0 +1,30 @@
+"""Tests for API view permissions in the irodsinfo app"""
+
+from django.urls import reverse
+
+# Projectroles dependency
+from projectroles.tests.test_permissions_api import TestProjectAPIPermissionBase
+
+
+class TestIrodsConfigRetrieveAPIView(TestProjectAPIPermissionBase):
```
mikkonie: Again, there is no project context so we shouldn't use the project testing base class or project users.

irodsinfo/views.py
```diff
@@ -22,6 +22,45 @@
 logger = logging.getLogger(__name__)
 
 
+class IrodsConfigMixin:
+    """Mixin for iRODS configuration views"""
+
+    @staticmethod
+    def create_env_config(user, irods_backend) -> dict:
```
mikkonie: Good work refactoring. I would name this method `get_irods_client_env()`.

Also, we do not use type hints in SODAR, as they kind of have to be used everywhere to work in a meaningful way. One day we'll do a huge refactoring and add them everywhere :)

mikkonie (changes requested):
Good work, had some minor notes, mostly on testing.

irodsinfo/views_api.py
```diff
@@ -0,0 +1,49 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class IrodsConfigRetrieveAPIView(IrodsConfigMixin, APIView):
+
+    """
+    Retrieve iRODS configuration file for the current user.
+
+    **URL:** ``/samplesheets/api/investigation/retrieve/{Project.sodar_uuid}``
```
mikkonie: This URL seems to be incorrect.

irodsinfo/views_api.py
```diff
@@ -0,0 +1,49 @@
...
+from projectroles.plugins import get_backend_api
+
+from irodsinfo.views import IrodsConfigMixin
+
+
+logger = logging.getLogger(__name__)
+
+
+class IrodsConfigRetrieveAPIView(IrodsConfigMixin, APIView):
+
```
mikkonie: Unnecessary empty line here :)

irodsinfo/views_api.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    Retrieve iRODS configuration file for the current user.
+
+    **URL:** ``/samplesheets/api/investigation/retrieve/{Project.sodar_uuid}``
+
+    **Methods:** ``GET``
+
+    **Returns:**
+
+    - ``irods_environment``: Return user iRODS config (``json``)
```
mikkonie: This should say `iRODS client environment`. JSON is a given, all responses are in JSON unless otherwise specified.

irodsinfo/views_api.py
```diff
@@ -0,0 +1,49 @@
...
+# Projectroles dependency
+from projectroles.plugins import get_backend_api
+
+from irodsinfo.views import IrodsConfigMixin
+
+
+logger = logging.getLogger(__name__)
+
+
+class IrodsConfigRetrieveAPIView(IrodsConfigMixin, APIView):
```
mikkonie: One naming convention related thing I missed earlier:

The original UI view returns both the iRODS client environment and a possible certificate file. We refer to the both of these as "iRODS configuration". However, this view only returns the client environment file and nothing else.

Hence, for clarity to the user, we should edit the view, url and docstring to reflect that only the environment is returned.

My suggestion:

- Rename this API view to `IrodsEnvRetrieveAPIView`
- Rename the URL name to `api_env`
- Change the URL pattern to `api/environment`
- Update docstring(s) to state we specifically return the iRODS client environment.

Sorry for not catching this earlier.

mikkonie (changes requested):

mikkonie (approved):
Looks good, aprpoved!

PR #1715, opened 2023-06-19
Landingzones status update

### LINKED ISSUE #1684: Query for UI landing zone status in batch
Currently, we do a query to `ZoneStatusRetrieveAjaxView` for each landing zone in the zone list UI. If we have a bunch of existing zones, this creates a flood of requests for the server, clogging up debug logs and this annoys me :)

Instead, we should do one periodic query for all zones in the list and update them.

## Spec

- Update `ZoneStatusRetrieveAjaxView` to take a `POST` request instead of `GET`
    * Replace `landingzone` kwarg in URL with `project`
- Receive list of `LandingZone` UUIDs in the `POST` data
- Check perms for each zone as we currently do
    * Also ensure zone belongs in `project` from URL kwarg
- Return dict of statuses by UUID
    * E.g. `{'zone-uuid-here': {'status': 'xxx', 'status_info': 'yyy'}`
- Update `landingzones.js`
    * Collect zone UUIDs from template
    * Call status view with list of UUIDs
    * Update zones with response data
- Ensure this also works in the project detail view
- Update tests as needed

gromdimon:
### Summary:
- Updated ZoneStatusRetrieveAjaxView to take a POST request instead of GET
- Updated landingzones.js
- Updated tests as needed
-  Added superuser variable in Jquery
- Tested


gromdimon:
General comment:
```
This works as specified, but one thing I didn't think of were the information security issues with passing info such as is_superuser to the client from the template.

These variables can be changed in run-time in the inspector, so technically any user could "gain" superuser access. In this particular case it's not super critical, but technically it's not something we should do.

I'd like you to try and refactor this in the following way:

In landingzones.js, once the document is ready, do an ajax call to CurrentUserRetrieveAjaxView
See SODAR Core for that view
Store the superuser status as a javascript variable from the view's response
..and, you guessed it, use that variable instead :)
This way the info can't be forged.
```
Link: https://github.com/bihealth/sodar-server/pull/1712#issuecomment-1594551129

mikkonie (approved):
Great work on this one, no notes :)

PR #1719, opened 2023-06-27
Output irodsorphans results step by step

### LINKED ISSUE #1319: Output irodsorphans command results as they are discovered
Instead of outputting all the results only at the end of execution, we should output each line as we encounter it, as is customary in other management commands.

The current implementation suffers if there is e.g. a crash in the middle of execution: all data gathered so far is for naught.

So, in brief:

## Spec

- Retain current functionality for the content of the output
- Print out each line of results as we enter them, instaed of printing out the whole list in the end.

### LINKED ISSUE #1741: Sort irodsorphans admin command output by Project.full_title
Sorting the project filtering by `Project.full_title` makes the results easier to parse. Can be done simultaneously with #1319.

gromdimon:
### TODO:
- [x] Implement
- [x] Test

### DIFF
--- a/irodsadmin/management/commands/irodsorphans.py
+++ b/irodsadmin/management/commands/irodsorphans.py
@@ -9,7 +9,7 @@ from django.template.defaultfilters import filesizeformat
 
 # Projectroles dependency
 from projectroles.management.logging import ManagementCommandLogger
-from projectroles.models import Project
+from projectroles.models import Project, PROJECT_TYPE_PROJECT
 from projectroles.plugins import get_backend_api
 
 # Landingzones dependency
@@ -38,6 +38,15 @@ class Command(BaseCommand):
     def __init__(self):
         super().__init__()
         self.irods_backend = get_backend_api('omics_irods')
+
+    def _get_uuid_from_path(self, path):
+        """Helper function to get UUID from a path"""
+        match = re.search(
+            r'/([a-f0-9]{2})/\1[a-f0-9]{6}-([a-f0-9]{4}-){3}[a-f0-9]{12}$', path
+        )
+        if match:
+            return match.group(1)
+        return None
 
     def _get_assay_collections(self, assays):
         """Return a list of all assay collection names."""
@@ -141,9 +150,31 @@ class Command(BaseCommand):
         Return a list of orphans in a given irods session that are not in a given
         list of expected collections.
         """
-        collections = irods.collections.get('/{}/projects'.format(irods.zone))
+        # Get a sorted list of all project collections
+        all_project_collections = sorted(
+            self.irods_backend.get_colls_recursively(
+                irods.collections.get('/{}/projects'.format(irods.zone))
+            ),
+            key=lambda coll: coll.path,
+        )
+        # Query for all projects of type PROJECT and store in a dict
+        project_dict = {
+            project.full_title: project.sodar_uuid
+            for project in Project.objects.filter(type=PROJECT_TYPE_PROJECT)
+        }
+        # Sort dict by full_title
+        project_dict = dict(
+            sorted(project_dict.items(), key=lambda item: item[0])
+        )
+        # Sort collections by project full_title
+        sorted_project_collections = sorted(
+            all_project_collections,
+            key=lambda coll: project_dict.get(
+                self._get_uuid_from_path(coll.path), float('inf')
+            ),
+        )
 
-        for collection in self.irods_backend.get_colls_recursively(collections):
+        for collection in sorted_project_collections:
             if (
                 self._is_zone(collection)
                 or self._is_assay_or_study(collection)
--- a/irodsadmin/management/commands/irodsorphans.py
+++ b/irodsadmin/management/commands/irodsorphans.py
@@ -3,6 +3,8 @@
 import re
 import sys
 
+from itertools import chain
+
 from django.core.management.base import BaseCommand
 from django.db.models import Q
 from django.template.defaultfilters import filesizeformat
@@ -62,7 +64,7 @@ class Command(BaseCommand):
             )
             match = re.search(r'{}'.format(pattern), coll.path)
             uuid = match.string.split('/')[4] if match else ''
-            if any(uuid in path for path in valid_project_paths):
+            if uuid and any(uuid in path for path in valid_project_paths):
                 colls_with_project.append(coll)
             else:
                 colls_no_project.append(coll)
@@ -165,12 +167,22 @@ class Command(BaseCommand):
 
     def _is_assay_or_study(self, collection):
         """
-        Check if a given collection matches the format of path to a study or assay
-        collection.
+        Check if a given collection matches the format of path to a study or
+        assay collection.
         """
         return re.match(
             r'(assay|study)_[a-f0-9]{8}-([a-f0-9]{4}-){3}[a-f0-9]{12}',
             collection.name,
+        )
+
+    def _is_assay_orphan(self, collection):
+        """
+        Check if a given collection matches the format of path to a study or
+        assay orphan.
+        """
+        return re.search(
+            r'(assay|study)_[a-f0-9]{8}-([a-f0-9]{4}-){3}[a-f0-9]{12}',
+            collection.path,
         )
 
     def _is_project(self, collection):
@@ -197,27 +209,30 @@ class Command(BaseCommand):
             ),
             key=lambda coll: coll.path,
         )
+        assay_collections = list(
+            chain.from_iterable(
+                self.irods_backend.get_child_colls(
+                    irods, self.irods_backend.get_path(a)
+                )
+                for a in assays
+                if a.get_plugin()
+            )
+        )
+
         # Sort collections by project full_title
-        sorted_collections = self._sort_colls_on_projects(project_collections)
+        sorted_collections = self._sort_colls_on_projects(
+            project_collections + assay_collections
+        )
 
         for collection in sorted_collections:
             if (
                 self._is_zone(collection)
                 or self._is_assay_or_study(collection)
                 or self._is_project(collection)
+                or self._is_assay_orphan(collection)
             ):
                 if collection.path not in expected:
                     self._write_orphan(collection.path, irods)
-
-        for assay in assays:
-            if not assay.get_plugin():
-                continue
-            with self.irods_backend.get_session() as irods:
-                for collection in self.irods_backend.get_child_colls(
-                    irods, self.irods_backend.get_path(assay)
-                ):
-                    if collection.path not in expected:
-                        self._write_orphan(collection.path, irods)
 
     def _write_orphan(self, path, irods):
         stats = self.irods_backend.get_object_stats(irods, path)
@@ -247,7 +262,7 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):
         studies = list(Study.objects.all())
-        assays = list(Assay.objects.all())
+        assays = list(Assay.objects.all().order_by())
         expected = (
             *self._get_assay_collections(assays),
             *self._get_study_collections(studies),
--- a/irodsadmin/management/commands/irodsorphans.py
+++ b/irodsadmin/management/commands/irodsorphans.py
@@ -1,6 +1,7 @@
 """Irodsorphans management command"""
 
 import re
+import sys
 
 from django.core.management.base import BaseCommand
 from django.db.models import Q
@@ -166,7 +167,7 @@ def log_orphan(orphan, irods_backend, irods):
     else:
         uuid = '<ERROR>'
         title = '<ERROR>'
-    logger.debug(
+    sys.stdout.write(
         ';'.join(
             [
                 uuid,
@@ -176,6 +177,7 @@ def log_orphan(orphan, irods_backend, irods):
                 filesizeformat(stats['total_size']).replace(u'\xa0', ' '),
             ]
         )
+        + '\n'
     )
 
 
--- a/irodsadmin/management/commands/irodsorphans.py
+++ b/irodsadmin/management/commands/irodsorphans.py
@@ -115,7 +115,13 @@ class Command(BaseCommand):
         Check if a given collection matches the format of path to a landing zone
         collection.
         """
-        return '/landing_zones/' in collection.path and re.search(
+        projects_path = self.irods_backend.get_projects_path()
+        pattern = (
+            projects_path
+            + r'/([a-f0-9]{2})/\1[a-f0-9]{6}-([a-f0-9]{4}-){3}[a-f0-9]{12}/'
+            r'landing_zones'
+        )
+        return re.search(r'{}'.format(pattern), collection.path) and re.search(
             r'^\d{8}_\d{6}', collection.name
         )
 
--- a/irodsadmin/management/commands/irodsorphans.py
+++ b/irodsadmin/management/commands/irodsorphans.py
@@ -178,13 +178,14 @@ class Command(BaseCommand):
 
         # Get the actual path to the projects collection
         project_path = self.irods_backend.get_projects_path()
+        depth = len(project_path.split('/')) + 1
         for coll in collections:
             pattern = (
                 project_path
                 + r'/([a-f0-9]{2})/\1[a-f0-9]{6}-([a-f0-9]{4}-){3}[a-f0-9]{12}'
             )
             match = re.search(r'{}'.format(pattern), coll.path)
-            uuid = match.string.split('/')[4] if match else ''
+            uuid = match.string.split('/')[depth] if match else ''
             if (
                 uuid
                 and any(uuid in path for path in valid_project_paths)
@@ -204,8 +205,8 @@ class Command(BaseCommand):
                     i
                     for i, path in enumerate(valid_project_paths)
                     if (
-                        coll.path.split('/')[4]
-                        if len(coll.path.split('/')) > 4
+                        coll.path.split('/')[depth]
+                        if len(coll.path.split('/')) > depth
                         else ''
                     )
                     in path
@@ -255,7 +256,9 @@ class Command(BaseCommand):
 
     def _write_orphan(self, path, irods):
         stats = self.irods_backend.get_object_stats(irods, path)
-        m = re.search(r'/projects/([^/]{2})/(\1[^/]+)', path)
+        projects_path = self.irods_backend.get_projects_path()
+        pattern = projects_path + r'/([^/]{2})/(\1[^/]+)'
+        m = re.search(pattern, path)
         if m:
             uuid = m.group(2)
             try:
--- a/irodsadmin/management/commands/irodsorphans.py
+++ b/irodsadmin/management/commands/irodsorphans.py
@@ -42,59 +42,6 @@ class Command(BaseCommand):
         super().__init__()
         self.irods_backend = get_backend_api('omics_irods')
 
-    def _sort_colls_on_projects(self, collections):
-        """Helper function to sort collections based on project list"""
-        colls_with_project = []
-        colls_no_project = []
-        temp_paths = []
-
-        # Create a set of valid project paths based on project UUIDs
-        valid_project_paths = [
-            self.irods_backend.get_path(p)
-            for p in Project.objects.filter(type=PROJECT_TYPE_PROJECT).order_by(
-                'full_title'
-            )
-        ]
-
-        # Get the actual path to the projects collection
-        project_path = self.irods_backend.get_projects_path()
-        for coll in collections:
-            pattern = (
-                project_path
-                + r'/([a-f0-9]{2})/\1[a-f0-9]{6}-([a-f0-9]{4}-){3}[a-f0-9]{12}'
-            )
-            match = re.search(r'{}'.format(pattern), coll.path)
-            uuid = match.string.split('/')[4] if match else ''
-            if (
-                uuid
-                and any(uuid in path for path in valid_project_paths)
-                and coll.path not in temp_paths
-            ):
-                colls_with_project.append(coll)
-                temp_paths.append(coll.path)
-            elif coll.path not in temp_paths:
-                colls_no_project.append(coll)
-                temp_paths.append(coll.path)
-
-        # Sort collections with project path based on project list
-        sorted_colls = sorted(
-            colls_with_project,
-            key=lambda coll: next(
-                (
-                    i
-                    for i, path in enumerate(valid_project_paths)
-                    if (
-                        coll.path.split('/')[4]
-                        if len(coll.path.split('/')) > 4
-                        else ''
-                    )
-                    in path
-                ),
-                float('inf'),
-            ),
-        )
-        return sorted_colls + colls_no_project
-
     def _get_assay_collections(self, assays):
         """Return a list of all assay collection names."""
         return [self.irods_backend.get_path(a) for a in assays]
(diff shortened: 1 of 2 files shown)


irodsadmin/management/commands/irodsorphans.py
```diff
@@ -151,37 +150,33 @@ def get_orphans(irods, irods_backend, expected, assays):
...
+        uuid = m.group(2)
+        try:
+            project = Project.objects.get(sodar_uuid=uuid)
+            title = project.full_title
+        except Project.DoesNotExist:
+            title = '<DELETED>'
+    else:
+        uuid = '<ERROR>'
+        title = '<ERROR>'
+    logger.debug(
```
mikkonie: Do not use logger here, use `self.stdout.write()` like in the original implementation.

Also note that if you'd use `logger.debug()`, we would never see anything in production where this admin command is actually used, because the log level is only debug when developing :)

mikkonie (changes requested):
Had one request for changes, other than that no comments. I'll check again after tests have been updated.

irodsadmin/tests/test_commands.py
```diff
@@ -206,49 +209,59 @@ def test_is_assay_or_study_invalid(self):
...
+        sys.stdout = out
+        irodsorphans.get_orphans(
+            self.irods,
+            self.irods_backend,
+            self.expected_collections,
+            [self.assay],
         )
+        output = out.getvalue()
+        sys.stdout = sys.__stdout__
+        self.assertIn(orphan_path, output)
```
gromdimon: Here is `assertIn` and not `assertEqual` due to some additional information in stdout. Here is an output:
```
    self.assertEqual(orphan_path, output)
AssertionError: '/sodarZone/projects/fa/fa2a0714-7c97-44b9[70 chars]3456' != 'fa2a0714-7c97-44b9-b762-1f370cac6828;Test[131 chars]es\n'
- /sodarZone/projects/fa/fa2a0714-7c97-44b9-b762-1f370cac6828/landing_zones/user/small_germline_study/20201031_123456
+ fa2a0714-7c97-44b9-b762-1f370cac6828;TestProject;/sodarZone/projects/fa/fa2a0714-7c97-44b9-b762-1f370cac6828/landing_zones/user/small_germline_study/20201031_123456;0;0 bytes
```

irodsadmin/tests/test_commands.py
```diff
@@ -278,55 +296,42 @@ def test_get_orphans_project(self):
...
-            irodsorphans.get_orphans(
-                self.irods,
-                self.irods_backend,
-                self.expected_collections,
-                [self.assay],
-            ),
-            [orphan_path],
-        )
 
-    def test_get_output(self):
```
gromdimon, line 308: In new implementation this test got redundant, as we combined producing and logging steps in one

irodsadmin/management/commands/irodsorphans.py
```diff
@@ -151,37 +151,34 @@ def get_orphans(irods, irods_backend, expected, assays):
...
-                    title,
-                    orphan,
-                    str(stats['file_count']),
-                    filesizeformat(stats['total_size']).replace(u'\xa0', ' '),
-                ]
-            )
+                    log_orphan(collection.path, irods_backend, irods)
+
+
+def log_orphan(orphan, irods_backend, irods):
```
mikkonie: These helpers should be refactored as protected methods within the command class.

That way, we can also simply set `self.irods_backend` in `__init__()` and eliminate the need to provide it to each helper function as an argument. The `irods` session object should still be given as arg though, because we want to init it in a `with` block in `handle()` as we currently do.

irodsadmin/tests/test_commands.py
mikkonie: Comparing the output to the path using `assertEqual()` will obviously not work. As you see from when you call `stdout.write`, we don't simply write the path buth the following values separated by semicolons:

- Project UUID (if project exists)
- Project title (if project exists)
- iRODS path
- File count in iRODS
- File size in iRODS

The tests should test this full output and not simply the occurrence of the path. In some existing tests this is actually done, see an example there :)

irodsadmin/management/commands/irodsorphans.py
```diff
@@ -151,37 +151,34 @@ def get_orphans(irods, irods_backend, expected, assays):
...
+    stats = irods_backend.get_object_stats(irods, orphan)
```
mikkonie: This is something left over from the original implementation, but I'd like a couple of additional refactorings to the `log_orphan()` method:

- Rename into `write_orphan()` (since we're not using logging here)
- Rename `orphan` argument into `path` ("orphan" is ambiguous and this is specifically the iRODS collection path)

irodsadmin/management/commands/irodsorphans.py
```diff
@@ -151,37 +151,34 @@ def get_orphans(irods, irods_backend, expected, assays):
...
+    m = re.search(r'/projects/([^/]{2})/(\1[^/]+)', orphan)
+    if m:
+        uuid = m.group(2)
+        try:
+            project = Project.objects.get(sodar_uuid=uuid)
+            title = project.full_title
+        except Project.DoesNotExist:
+            title = '<DELETED>'
```
mikkonie: These `<DELETED>` and `<ERROR>` strings could be changed into constants to keep this constant (pun intended) and make testing more straightforward.

irodsadmin/tests/test_commands.py
```diff
@@ -437,13 +463,17 @@ def test_command_orphan_assay_sub(self):
         )
         self.irods.collections.create(orphan_path)
         out = io.StringIO()
+        sys.stdout = out
         call_command('irodsorphans', stdout=out)
+        output = out.getvalue()
```
mikkonie: The following bit seems to be repeated quite a lot in the tests without change:

```
out = io.StringIO()
sys.stdout = out
call_command('irodsorphans', stdout=out)
output = out.getvalue()
```

How about making this a protected helper method in the test class which returns `output`?

mikkonie (changes requested):
This works as expected, good work. I had some requests for refactoring.

irodsadmin/management/commands/irodsorphans.py
```diff
@@ -24,184 +25,175 @@
...
-            *get_assay_collections(assays, irods_backend),
-            *get_study_collections(studies, irods_backend),
-            *get_zone_collections(irods_backend),
-            *get_project_collections(irods_backend),
-            *get_assay_subcollections(studies, irods_backend),
+            *self._get_assay_collections(assays),
+            *self._get_study_collections(studies),
+            *self._get_zone_collections(),
+            *self._get_project_collections(),
+            *self._get_assay_subcollections(studies),
```
mikkonie, line 297: While you were on vacation, I added issue #1741 with the suggestion it would be done together with this task.

This will require some reactoring:

- Iterate through projects (sorted by `full_title`)
- For each project, get expected collections
- For maximum clarity, for each project sort expected collections within that project before adding to the list

This should ensure the results are displayed in an order sorted by project. Tests should be updated to ensure this of course.

irodsadmin/tests/test_commands.py
```diff
@@ -114,45 +111,49 @@ def tearDown(self):
         self.irods.cleanup()
         super().tearDown()
 
+    @staticmethod
+    def catch_stdout():
+        """Catch stdout from 'irodsorphans' management command"""
```
mikkonie: Minor style note, we don't need the hyphens around the command name here :)

mikkonie (changes requested):
Looking good, I'd like you to incorporate #1741 into this pull request.

irodsadmin/management/commands/irodsorphans.py
gromdimon: Hmm, I don't seem to get the point. I pushed through the implementation where we sort projects by full title, but I'm not sure if that's exactly what you asked for.

In the current implementation, the orphan registration ordering does not depend on the `expected` list, but on the `collections`, which we get with `irods.collections.get('/{}/projects'.format(irods.zone) )`. If we want to change the order of the output, then we must filter this list of `collections`, which is not a trivial task since there are deleted projects (without full_title) or assays that are linked to projects through studies and investigations.

Could you clarify what exactly is needed, because I'm afraid I didn't get the point of the task...

mikkonie: Yes, you are correct that the current implementation does not iterate by project but by the iRODS collection and I would like to change that. Right now, you can have the same project referred to in the output twice in different locations, due to first checking for orphaned landing zones, then for orphaned assay collections, etc.

A possible simple fix for this would be the following:

- After querying for all project collections with `get_colls_recursively()` in `_get_orphans()`, sort them by full path
    * This will already sort them by project, although not necessarily in the correct order by full project title
- How to make this even nicer?
    * Query for all projects of type PROJECT, store in dict as `{sodar_uuid: full_title}`
    * Compare iRODS paths to project UUID
    * Sort collections by project full_title based on the dict
    * Then continue onwards with evaluating collections

At least the first step is definitely needed, but the second one would be super nice to have, as it eliminates the manual work of sorting the output in e.g. Excel.

irodsadmin/management/commands/irodsorphans.py
```diff
@@ -25,184 +26,224 @@
...
+    def _sort_colls_on_projects(self, all_project_collections, project_list):
+        """Helper function to sort collections based on project list"""
+        # Separate strings with UUIDs from those without
+        strings_with_uuids = []
+        strings_without_uuids = []
+
+        # Iterate through L1 and classify strings
+        for coll in all_project_collections:
+            match = re.search(
+                r'[a-f0-9]{8}-([a-f0-9]{4}-){3}[a-f0-9]{12}', coll.path
```
mikkonie: Note that simply searching for the UUID is not OK in this case. Nothing is preventing a user from creating a collection to a landing zone with the name of an UUID of some other project. We need to check that the project UUID appears at the correct place in the path, namely after `{projects_path}/{two_first_characters_of_uuid}/{PROJECT_UUID_GOES_HERE}`. This we already too e.g. when validating iRODS delete requests.

One possible simple way would be to call `IrodsAPI.get_path(project)` for each project when we get the project list and store the project path in the sorted list. Then in here we could simply check if the start of the path is found with `any(startswith())` etc. But there might be an even better approach, feel free to give one a try :)

irodsadmin/management/commands/irodsorphans.py
```diff
@@ -25,184 +26,224 @@
...
-                    orphans.append(collection.path)
-    return orphans
 
+    def _get_orphans(self, irods, expected, assays):
+        """
+        Return a list of orphans in a given irods session that are not in a given
+        list of expected collections.
+        """
+        # Get a sorted list of all project collections
+        all_project_collections = sorted(
```
mikkonie: Especially for internal methods, we don't generally want to use excessively long variable names.

irodsadmin/management/commands/irodsorphans.py
```diff
@@ -25,184 +26,224 @@
...
+            self.irods_backend.get_colls_recursively(
+                irods.collections.get('/{}/projects'.format(irods.zone))
+            ),
+            key=lambda coll: coll.path,
+        )
+        # Query for all projects of type PROJECT and store in a dict
+        project_dict = {
+            project.full_title: project.sodar_uuid
+            for project in Project.objects.filter(type=PROJECT_TYPE_PROJECT)
```
mikkonie: Always do your sorting already in the Django database query if possible. That will perform much better when querying for large amounts of data. E.g. here just do `.order_by('full_title')`.

irodsadmin/management/commands/irodsorphans.py
```diff
@@ -25,184 +26,224 @@
...
+        }
+        # Sort dict by full_title and extract the UUID values
+        project_list = [str(val) for _, val in sorted(project_dict.items())]
```
mikkonie: I''m not sure why this line of code is needed, if we don't use `project_dict` for anything else. You can sort the objects in the query and grab just the UUID into a list, even using `values_list('sodar_uuid', flat=True`) to grab simply the UUID. Thus there should not be a need to iterate through the query results twice.

irodsadmin/tests/test_commands.py
```diff
@@ -471,4 +497,34 @@ def test_command_multiple(self):
             self.project.full_title,
             orphan_path,
         )
-        self.assertEqual(expected, out.getvalue())
+        self.assertEqual(expected, output)
+
+    def test_command_ordering(self):
+        """Test ordering of orphans in command output"""
+        project1 = self.make_project('A_Project', PROJECT_TYPE_PROJECT, None)
+        project2 = self.make_project('B_Project', PROJECT_TYPE_PROJECT, None)
```
mikkonie: This tests looks good, but one small thing: Since we already create `self.project` in `setUp()`, we only need to create one addlitional project here within the test case. Otherwise the original project is needlessly used and we waste many precicious CPU cycles ;)

mikkonie (changes requested):
Reviewed, added some suggestions on how to improve the ordering.

irodsadmin/management/commands/irodsorphans.py
```diff
@@ -25,184 +26,226 @@
...
+        valid_project_paths = [
+            self.irods_backend.get_path(p)
+            for p in Project.objects.filter(type=PROJECT_TYPE_PROJECT).order_by(
+                'full_title'
+            )
+        ]
+
+        for coll in collections:
+            uuid_prefix = (
+                coll.path.split('/')[4] if len(coll.path.split('/')) > 4 else ''
```
mikkonie: Here, note that we don't necessarily now that the project collection is this deep in the iRODS path. Right now in production and in our local dev setup, the projects collection is in `/sodarZone/projects`, but this may not always be the case (and, indeed, we are likely to change that in production at some point). 

You can get the actual position of the UUID from e.g. calling `IrodsAPI.get_projects_path()` and calculating the depth from there.

irodsadmin/management/commands/irodsorphans.py
```diff
@@ -25,184 +26,226 @@
...
+        # Create a set of valid project paths based on project UUIDs
+        valid_project_paths = [
+            self.irods_backend.get_path(p)
+            for p in Project.objects.filter(type=PROJECT_TYPE_PROJECT).order_by(
+                'full_title'
+            )
+        ]
+
+        for coll in collections:
+            uuid_prefix = (
```
mikkonie: Minor note but this is not a "UUID prefix", it's the entire UUID. It had me confused there for a while :)

irodsadmin/management/commands/irodsorphans.py
```diff
@@ -25,184 +26,226 @@
...
-                    orphans.append(collection.path)
-    return orphans
 
+    def _is_project(self, collection):
+        """
+        Check if a given collection matches the format of path to a project
+        collection.
+        """
+        return re.search(
+            r'projects/([a-f0-9]{2})/\1[a-f0-9]{6}-([a-f0-9]{4}-){3}[a-f0-9]{12}$',
```
mikkonie: Similar to my other comment, we should be ensuring the entire start of the path instead of just looking "projects" or the UUID somewhere within the path, as users can technically create arbitrarily named collections within the path.

The regexes should search for things under the projects path retrieved with `IrodsAPI.get_projects_path()`, adding the begining of the line character `^` to make sure we are evaluating the correct part of the path. Maybe this was already wrong in the original implementation, in that case I've simply missed it before.

mikkonie (changes requested):
Reviewed, I had some requests for changes.

irodsadmin/management/commands/irodsorphans.py
```diff
@@ -25,184 +26,234 @@
...
+class Command(BaseCommand):
+    """Command to find orphans in iRODS collections."""
+
+    help = 'Find orphans in iRODS project collections.'
+
+    def __init__(self):
+        super().__init__()
+        self.irods_backend = get_backend_api('omics_irods')
+
+    def _sort_colls_on_projects(self, collections):
```
mikkonie: Sorting collections by projects doesn't seem to be working correctly.

In my dev project, I have two projects with both orphaned sample repository data and landing zone data. One of the projects has been deleted while the other still exists.

In the output below, you can see that the data is still ordered so that landing zones from all projects come first, followed by sample data from all projects. All of this should be sorted so that:

- We first get all from project X
- Then we get all from project Y

Output:
```
6fcdf2b7-b4fe-44d8-a6fe-0634eba54b7d;Test Category / Orphaned Data Project;/sodarZone/projects/6f/6fcdf2b7-b4fe-44d8-a6fe-0634eba54b7d/landing_zones/alice/demo_study/demo_assay/20230818_124813_orphan_zone;0;0 bytes
43fab04b-4bbe-4b1e-9ee9-a9d0dd4abb09;<DELETED>;/sodarZone/projects/43/43fab04b-4bbe-4b1e-9ee9-a9d0dd4abb09;2;40 bytes
43fab04b-4bbe-4b1e-9ee9-a9d0dd4abb09;<DELETED>;/sodarZone/projects/43/43fab04b-4bbe-4b1e-9ee9-a9d0dd4abb09/landing_zones/alice/investigation_title/investigation_title_transcription_profiling_nucleotide_sequencing/20230719_153525_orphaned_zone;1;20 bytes
43fab04b-4bbe-4b1e-9ee9-a9d0dd4abb09;<DELETED>;/sodarZone/projects/43/43fab04b-4bbe-4b1e-9ee9-a9d0dd4abb09/sample_data/study_956c34ad-4f54-4770-9fa5-0d8c720336e5;1;20 bytes
43fab04b-4bbe-4b1e-9ee9-a9d0dd4abb09;<DELETED>;/sodarZone/projects/43/43fab04b-4bbe-4b1e-9ee9-a9d0dd4abb09/sample_data/study_956c34ad-4f54-4770-9fa5-0d8c720336e5/assay_a333c690-c95e-4a1e-be0e-2f7bef5c2615;1;20 bytes
6fcdf2b7-b4fe-44d8-a6fe-0634eba54b7d;Test Category / Orphaned Data Project;/sodarZone/projects/6f/6fcdf2b7-b4fe-44d8-a6fe-0634eba54b7d/sample_data/study_845d4d27-47d9-4273-9fea-62cf55b5de79/assay_ea75df18-2911-456b-8412-06bb1df4607d/orphaned_collection;2;45 bytes
```

mikkonie (changes requested):

mikkonie (changes requested):
The ordering is correct now. However, there is sadly another bug. For assay paths, non-orphaned paths get reported in the output.

In my test project, I have an assay collection of `P1001-N1-DNA1-WGS1`, which is correctly discovered by the corresponding assay plugin. That path is not reported as an orphan, which is correct.

However, under that collection I have some subcollections. Those get reported by the admin command, even though they are of course not orphaned (their parent collection is very much correct).

Before this didn't happen, so some change has recently caused this. My guess is a comparison which before was "path starts with X" is now "path is exactly X".

irodsadmin/management/commands/irodsorphans.py
```diff
@@ -25,184 +28,254 @@
...
+            self.irods_backend.get_path(p)
+            for p in Project.objects.all().order_by('full_title')
+        ]
+
+    def _is_zone(self, collection):
+        """
+        Check if a given collection matches the format of path to a landing zone
+        collection.
+        """
+        return '/landing_zones/' in collection.path and re.search(
```
mikkonie: This check also looks for a substring, which is not a reliable way to ensure a collection is what we expect since users can create arbitrarily named collections in iRODS. We should ensure the path starts with the path to a project and then is followed by `/landing_zones/` etc, as we do now.

And I know, this is a problem with the original implementation, but let's fix everything while we're at it :)

irodsadmin/management/commands/irodsorphans.py
```diff
@@ -25,184 +28,254 @@
...
-        for assay in study.assays.all():
-            assay_table = study_tables['assays'][str(assay.sodar_uuid)]
-            assay_plugin = assay.get_plugin()
-            assay_path = irods_backend.get_path(assay)
+    def _is_assay_or_study(self, collection):
+        """
+        Check if a given collection matches the format of path to a study or
+        assay collection.
+        """
+        return re.match(
```
mikkonie: Same thing here, we should ensure the start of the path is directly within a project's collection in iRODS.

irodsadmin/management/commands/irodsorphans.py
```diff
@@ -25,184 +28,254 @@
...
+            r'(assay|study)_[a-f0-9]{8}-([a-f0-9]{4}-){3}[a-f0-9]{12}',
+            collection.path,
+        )
+
+    def _is_project(self, collection):
+        """
+        Check if a given collection matches the format of path to a project
+        collection under the projects path.
+        """
+        project_path = self.irods_backend.get_projects_path()
```
mikkonie: Since the projects path is always the same, we can generate it once into a variable and then pass it to these helpers when iterating.

Oh and this should really be called `projects_path` since it's the root path for all projects, not just one.

irodsadmin/management/commands/irodsorphans.py
```diff
@@ -25,184 +28,254 @@
...
-    return [
-        irods_backend.get_path(lz)
-        for lz in LandingZone.objects.filter(
-            ~(Q(status=ZONE_STATUS_MOVED) & Q(status=ZONE_STATUS_DELETED))
+    def _is_assay_orphan(self, collection):
+        """
+        Check if a given collection matches the format of path to a study or
+        assay orphan.
+        """
+        return re.search(
```
mikkonie: Here we also need to check for the beginning of the path to ensure it's the expected one.

mikkonie (changes requested):
This works fine now, good work! I have a few requests to improve the regex checks.

irodsadmin/management/commands/irodsorphans.py
```diff
@@ -25,184 +28,266 @@
...
+
+        # Sort collections with project path based on project list
+        sorted_colls = sorted(
+            colls_with_project,
+            key=lambda coll: next(
+                (
+                    i
+                    for i, path in enumerate(valid_project_paths)
+                    if (
+                        coll.path.split('/')[4]
```
mikkonie: It seems we still use index here to determine the position of the project UUID. As I told before, this is not constant: right now in the dev environment projects go directly under `/sodarZone`, but this may not be the case.

The position of the UUID should instead be calculated by the depth of `IrodsAPI.get_projects_path()`. I already requested this change once, but maybe it was for some other place in the implementation. Make sure we are not assuming the UUID position with a fixed collection depth index anywhere.

irodsadmin/management/commands/irodsorphans.py
```diff
@@ -25,184 +28,266 @@
...
+                self._is_zone(collection)
+                or self._is_assay_or_study(collection)
+                or self._is_project(projects_path, collection)
+                or collection.path in assay_coll_paths
+            ) and collection.path not in expected:
+                self._write_orphan(collection.path, irods)
+
+    def _write_orphan(self, path, irods):
+        stats = self.irods_backend.get_object_stats(irods, path)
+        m = re.search(r'/projects/([^/]{2})/(\1[^/]+)', path)
```
mikkonie: As you updated elsewhere, here we also need to make sure we are searching for the beginning of the path, which needs to correspond to `get_projects_path()`.

mikkonie (changes requested):
Good work, I still found a few small things to complain about though :)

irodsadmin/management/commands/irodsorphans.py
```diff
@@ -25,184 +28,269 @@
...
+        ]
+
+    def _is_zone(self, collection):
+        """
+        Check if a given collection matches the format of path to a landing zone
+        collection.
+        """
+        projects_path = self.irods_backend.get_projects_path()
+        pattern = (
+            projects_path
```
mikkonie: I think there might still be a problem here. We don't provide the start of line character `^` to the regex. This still gives a chance we have a false match if someone creates collections with the same name somewhere in the path. Please also check the other regexes for this.

irodsadmin/tests/test_commands.py
```diff
@@ -471,4 +500,34 @@ def test_command_multiple(self):
...
+        self.make_assignment(project1, self.user, self.role_owner)
+        orphan_path1 = '{}/sample_data/study_{}'.format(
+            self.irods_backend.get_path(project1), str(uuid.uuid4())
+        )
+        # As the title of self.project is 'Test Project', it should be ordered
+        # after project1 with title 'A_Project'
+        orphan_path2 = '{}/sample_data/study_{}'.format(
+            self.irods_backend.get_path(self.project), str(uuid.uuid4())
+        )
+        self.irods.collections.create(orphan_path1)
```
mikkonie: Generally we want to label multiple variables from the second one onwards, so the first one would just be `orphan_path`. This creates consistency and helps searching for the variables for when we e.g. only have a single such variable in other cases.

mikkonie (changes requested):
Reviewed. Still some minor regex optimization to be done and a small style issue, otherwise looking good to merge.

irodsadmin/management/commands/irodsorphans.py
```diff
@@ -25,184 +28,273 @@
...
+                'full_title'
+            )
+        ]
+
+        # Get the actual path to the projects collection
+        project_path = self.irods_backend.get_projects_path()
+        depth = len(project_path.split('/')) + 1
+        for coll in collections:
+            pattern = (
+                project_path
```
mikkonie: Here the `^` regex character is still missing.

irodsadmin/tests/test_commands.py
mikkonie: This seems to be unaddressed from the previous review.

irodsadmin/management/commands/irodsorphans.py
```diff
@@ -25,184 +28,273 @@
...
+
+    def _get_zone_collections(self):
+        """
+        Return a list of all landing zone collection names that are not MOVED or
+        DELETED.
+        """
+        return [
+            self.irods_backend.get_path(lz)
+            for lz in LandingZone.objects.filter(
+                ~(Q(status=ZONE_STATUS_MOVED) & Q(status=ZONE_STATUS_DELETED))
```
mikkonie: I know this is from the original implementation, but I believe this could be more simply done with `LandingZone.objects.exclude(status__in=[ZONE_STATUS_MOVED, ZONE_STATUS_ACTIVE])`.

mikkonie (changes requested):
Good work, two small comments + one unhandled comment from previous review. I'm sure we can merge after the next one! :)

irodsadmin/tests/test_commands.py
gromdimon: But I've changed `orhpan_path1` to `orphan_path` and `orphan_path2` to `orphan_path1`. Could you check it one more time?

mikkonie: It should be `orphan_path` and `orphan_path2`. Sorry if I was unclear earlier.

mikkonie (changes requested):
One more change and we should be done!

mikkonie (approved):
All is looking OK, good work! This will be merged.

PR #1727, opened 2023-07-13
Display iRODS file count in sheet overview tab

### LINKED ISSUE #1295: Display iRODS file count in sheet overview tab
Instead of simply `Available` for the iRODS repository, we should include the `IrodsStatsBadge` to show the file statistics for the entire investigation. This would be consistent with the rest of the UI.

Need to add the sample data path to `sodarContext` on the server side to make this work.

## Spec

- Add `irods_path` to `SheetContextAjaxView` output
    * Retrieve using `irods_backend.get_sample_path(project)
    * Include this only if investigation exists, e.g. in `inv_data` within the `if inv:` block
    * If `not inv.irods_status`, set value to `None`
- Update Django tests
    * Add value in tests found `samplesheets.tests.test_views_ajax.TestSheetContextAjaxView`
- In `Overview.vue`, add `IrodsStatsBadge` and retrieve data
    * If `sodarContext.irods_status`, display the badge
    * Replace the existing "Available" badge with the `IrodsStatsBadge` component
    	+ Provide `sodarContext.irods_path` as the `irodsPath` property
        + See `SheetTableHeader` for example of `IrodsStatsBadge` inclusion
    * The "Not created" badge should remain as is, if `!irods_status`
- Update Vue app tests
    * Add value to vue app example test data `sodarContext.json` in the correct position
        + Nothing should be added to `sodarContextNoSheet.json`, because the investigation doesn't exist in that example
    * Add check for `IrodsStatsBadge` in `Overview.spec.js`
    	+ Again, see `SheetTableHeader.spec.js` for an example
    	+ A similar simple `exists()` check on a mocked component is OK, as these are just unit tests and we don't perform requests

gromdimon:
### TODO:
- [x] Implement
- [x] Test

### DIFF
--- a/samplesheets/vueapp/package-lock.json
+++ b/samplesheets/vueapp/package-lock.json
@@ -38,7 +38,6 @@
         "eslint-plugin-vue": "^7.20.0",
         "fetch-mock-jest": "^1.5.1",
         "jest": "^27.1.0",
-        "node-fetch": "^3.3.2",
         "vue-property-decorator": "^9.1.2",
         "vue-template-compiler": "^2.6.14",
         "webpack-bundle-tracker": "^1.4.0"
@@ -2679,25 +2678,6 @@
         "node": ">=8"
       }
     },
-    "node_modules/@nuxt/opencollective/node_modules/node-fetch": {
-      "version": "2.6.12",
-      "resolved": "https://registry.npmjs.org/node-fetch/-/node-fetch-2.6.12.tgz",
-      "integrity": "sha512-C/fGU2E8ToujUivIO0H+tpQ6HWo4eEmchoPIoXtxCrVghxdKq+QOHqEZW7tuP3KlV3bC8FRMO5nMCC7Zm1VP6g==",
-      "dependencies": {
-        "whatwg-url": "^5.0.0"
-      },
-      "engines": {
-        "node": "4.x || >=6.0.0"
-      },
-      "peerDependencies": {
-        "encoding": "^0.1.0"
-      },
-      "peerDependenciesMeta": {
-        "encoding": {
-          "optional": true
-        }
-      }
-    },
     "node_modules/@nuxt/opencollective/node_modules/supports-color": {
       "version": "7.2.0",
       "resolved": "https://registry.npmjs.org/supports-color/-/supports-color-7.2.0.tgz",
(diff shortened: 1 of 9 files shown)


samplesheets/vueapp/src/components/Overview.vue
```diff
@@ -33,15 +33,16 @@
...
           <dd class="col-md-9">
-            <span v-if="sodarContext.irods_status"
-                  class="badge badge-pill badge-success">
-              Available
+            <span v-if="sodarContext.irods_status">
+              <IrodsStatsBadge
+                :projectUuid="sodarContext.project_uuid"
+                :irodsStatus="sodarContext.irods_status"
+                :irodsPath="sodarContext.irods_path"
+              />
```
mikkonie: Minor style point, but please don't leave the closing part of the hanging, instead place it in the end of the previous row.

samplesheets/vueapp/src/components/Overview.vue
```diff
@@ -33,15 +33,16 @@
         <dl class="row pb-0">
           <dt class="col-md-3">iRODS Repository</dt>
           <dd class="col-md-9">
-            <span v-if="sodarContext.irods_status"
-                  class="badge badge-pill badge-success">
-              Available
+            <span v-if="sodarContext.irods_status">
```
mikkonie, line 36: Unfortunately this doesn't work: the badge remains in the "Updating.." phase forever, see accompanying screenshot.

No ajax request is received by the server either and upon closer inspection, it seems the `updateStats()` method is never called. See `SheetTableHeader` for an example on how to do this.

Also, as with other features, make sure to verify in your dev environment that this works as expected before sumitting the PR for review :)

2023-07_inv_badge_bug

mikkonie (changes requested):

samplesheets/vueapp/src/components/Overview.vue
mikkonie: This works now, good work fixing the issue.

samplesheets/vueapp/package-lock.json
```diff
@@ -38,6 +38,7 @@
         "eslint-plugin-vue": "^7.20.0",
         "fetch-mock-jest": "^1.5.1",
         "jest": "^27.1.0",
+        "node-fetch": "^3.3.2",
```
mikkonie: This is not needed, and package-lock.json should not be modified for this pull request. Please revert to the original packages.

samplesheets/vueapp/src/components/Overview.vue
```diff
@@ -33,15 +33,15 @@
...
+              <IrodsStatsBadge
+                :projectUuid="sodarContext.project_uuid"
+                :irodsStatus="sodarContext.irods_status"
+                :irodsPath="sodarContext.irods_path"/>
```
mikkonie: If you use the `/>` syntax, make sure to include a space before it.

samplesheets/vueapp/tests/unit/IrodsStatsBadge.spec.js
```diff
@@ -2,13 +2,17 @@ import { createLocalVue, mount } from '@vue/test-utils'
 import { projectUuid } from '../testUtils.js'
 import IrodsStatsBadge from '@/components/IrodsStatsBadge.vue'
 import '@/filters/prettyBytes.js'
+import fetch from 'node-fetch'
```
mikkonie: This should not be used, `node-fetch` breaks stuff (see e.g. the currently breaking CI tests) and it is not needed. Instead, just replace the `updateStats()` method with a stub. See `SheetTableHeader.spec.js` for how the setup is done there.

mikkonie (changes requested):
Reviewed, the functionality works now so good work. I did have some requests for changes.

mikkonie (approved):
Looks good, will be merged.

PR #1728, opened 2023-07-13
Add "create" tag for sheet versions & Add tests for templatetags

### LINKED ISSUE #1296: Add "create" tag for sheet versions
Currently, if we create sheets from a template we still end up with an "import" tag, which is technically wrong. Should use a "create" tag instead.

Remember to also update the docs regarding this.

## Spec

- Add `from_template` argument to `SampleSheetIO.import_isa()`
    * Boolean, default=`False`
- Update `SheetTemplateCreateForm.save()`
    * Set `from_template` to `True` in `import_isa()` call
- If `from_template=True`, set `tags = ['CREATE']`
    * Else `tags = ['IMPORT']` (the current implementation)
- Update samplesheets template tags
    * In `TAG_COLORS`, add `"CREATE": "info"`
- Update tests
    * In `TestSheetTemplateCreateFormView.test_post_batch()`, add check for `ISATab` object and that its `tags` value is `['CREATE']`
        + No need to do this in loop for every template, once after the loop on `ISATab.objects.first()` should be enough
    * In UI tests `TestSheetVersionListView`, test for import badge visibility
        + No need to test CREATE badge here since the code is so simple
    * (We should add template tag tests for samplesheets, but that's something for a different issue...)
- Cleanup of existing code
    * Rename `TestSheetTemplateCreateFormView` to `TestSheetTemplateCreateView`
    * Delete unused `ISATAB_TAGS` from `samplesheets.models`

### LINKED ISSUE #1723: Add samplesheets template tag tests
While writing the spec for #1296 I noticed these tests are completely missing.

For an example on how these are usually implemented, see e.g. `projectroles.tests.test_templatetags` in SODAR Core.

None of these tests should require taskflow or setting up the project in iRODS.

Low priority, this can be done after other tasks if running out of work :)

- Tests for `get_investigation()`
    * Test with no imported investigation
    * Test with imported investigation (use `import_isa_from_file()`, see existing tests) 
- Tests for `get_search_item_type()`
    * Test with `GENERIC_MATERIAL_TYPES` items
    * Test with special case of "file"
        + The values should be updated into constants
- Tests for `get_irods_tree()`
    * No need to test the entire formed HTML
    * Test for the existence of `IRODS_SAMPLE_COLL`, study and assay paths in the returned string
    * NOTE: Even though we use `irods_backend` for this, no taskflow or iRODS project setup is needed! (We just build paths in the irodsbackend API)
- Tests for `get_material_search_url()`
    * See in code how `item` is formed
    * Import ISA-Tab in test setup, provide study for the method
    * Assert URL is as expected
- Tests for `get_irods_path()`
    * Test with a couple of different object types (e.g. project and assay) and `sub_path`
- Tests for `get_icon()`
    * Test for study and assay
    * No need to assert the whole string, but test the expected `text-*` class and icon class are visible
- Tests for `get_isatab_tag_html()`
    * Loop test with tags in `TAG_COLORS`
    * Test with an unknown tag to assert `DEFAULT_TAG_COLOR` is returned
- Tests for `get_request_status_class()`
    * Test with values in `REQUEST_STATUS_CLASSES`
    * Test with unknown value
- Tests for `trim_base_path()`
    * Test with a realistic iRODS path (doesn't have to correspond to anything actually in iRODS)

gromdimon:
### TODO:
- [x] Implement
- [x] Test
- [x] Cleanup
- [x] Docs

### DIFF
--- a/samplesheets/tests/test_templatetags.py
+++ b/samplesheets/tests/test_templatetags.py
@@ -90,7 +90,7 @@ class TestSamplesheetsTemplateTags(
             comments=DEFAULT_COMMENTS,
         )
         # Setup iRODS backend for the test
-        self.irods_backend = get_backend_api("omics_irods")
+        self.irods_backend = get_backend_api('omics_irods')
 
     def test_get_investigation(self):
         """Test get_investigation()"""
@@ -106,14 +106,14 @@ class TestSamplesheetsTemplateTags(
     def test_get_search_item_type_with_material_types(self):
         """Test get_search_item_type() with material types"""
         for material_type in GENERIC_MATERIAL_TYPES:
-            item = {"type": material_type}
+            item = {'type': material_type}
             expected_type = GENERIC_MATERIAL_TYPES[material_type]
             self.assertEqual(s_tags.get_search_item_type(item), expected_type)
 
     def test_get_search_item_type_with_file(self):
-        """Test get_search_item_type() with special case "file" """
-        item = {"type": "file"}
-        self.assertEqual(s_tags.get_search_item_type(item), "Data File")
+        """Test get_search_item_type() with special case 'file'"""
+        item = {'type': 'file'}
+        self.assertEqual(s_tags.get_search_item_type(item), 'Data File')
 
     def test_get_irods_tree(self):
         """Test get_irods_tree()"""
@@ -124,23 +124,23 @@ class TestSamplesheetsTemplateTags(
         self.assertIn(self.irods_backend.get_sub_path(self.study), ret)
         # Assert that assay path exists in the returned string
         clean_assay_path_0 = self.irods_backend.get_sub_path(self.assay).split(
-            "/"
+            '/'
         )[0]
         clean_assay_path_1 = self.irods_backend.get_sub_path(self.assay).split(
-            "/"
+            '/'
         )[1]
         self.assertIn(clean_assay_path_0, ret)
         self.assertIn(clean_assay_path_1, ret)
 
     def test_get_material_search_url(self):
         """Test get_material_search_url()"""
-        item = {"study": self.study, "name": "Sample1"}
+        item = {'study': self.study, 'name': 'Sample1'}
         url = s_tags.get_material_search_url(item)
         expected_url = reverse(
-            "samplesheets:project_sheets",
-            kwargs={"project": self.project.sodar_uuid},
+            'samplesheets:project_sheets',
+            kwargs={'project': self.project.sodar_uuid},
         )
-        expected_url += "#/study/{}/filter/Sample1".format(
+        expected_url += '#/study/{}/filter/Sample1'.format(
             self.study.sodar_uuid
         )
         self.assertEqual(url, expected_url)
@@ -158,8 +158,8 @@ class TestSamplesheetsTemplateTags(
     def test_get_irods_path_with_project_and_sub_path(self):
         """Test get_irods_path() with project and sub_path"""
         project_path = self.irods_backend.get_path(self.project)
-        sub_path = "subfolder1/subfolder2"
-        expected_path = project_path + "/" + sub_path
+        sub_path = 'subfolder1/subfolder2'
+        expected_path = project_path + '/' + sub_path
         self.assertEqual(
             s_tags.get_irods_path(self.project, sub_path), expected_path
         )
@@ -167,8 +167,8 @@ class TestSamplesheetsTemplateTags(
     def test_get_irods_path_with_assay_and_sub_path(self):
         """Test get_irods_path() with assay and sub_path"""
         assay_path = self.irods_backend.get_path(self.assay)
-        sub_path = "subfolder1/subfolder2"
-        expected_path = assay_path + "/" + sub_path
+        sub_path = 'subfolder1/subfolder2'
+        expected_path = assay_path + '/' + sub_path
         self.assertEqual(
             s_tags.get_irods_path(self.assay, sub_path), expected_path
         )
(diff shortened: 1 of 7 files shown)


gromdimon:
## Another Issue: #1723 
### TODO:
- [x] Implement test


btw: I found an interesting approach to avoid imports of IrodsRequest and ISATab - just mocking them. I hope it works fine, if not - let me know and I will reimplement them.

samplesheets/tests/test_templatetags.py
```diff
@@ -0,0 +1,227 @@
...
+            file_name=ASSAY_FILE_NAME,
+            study=self.study,
+            tech_platform=ASSAY_TECH_PLATFORM,
+            tech_type=ASSAY_TECH_TYPE,
+            measurement_type=ASSAY_MEASURE_TYPE,
+            arcs=[],
+            comments=DEFAULT_COMMENTS,
+        )
+        # Setup iRODS backend for the test
+        self.irods_backend = get_backend_api("omics_irods")
```
mikkonie: Double quotes spotted! :)

samplesheets/tests/test_templatetags.py
```diff
@@ -0,0 +1,227 @@
...
+
+    def test_get_investigation_no_investigation(self):
+        """Test get_investigation() without investigation"""
+        self.investigation.delete()
+        self.assertEqual(s_tags.get_investigation(self.project), None)
+
+    def test_get_search_item_type_with_material_types(self):
+        """Test get_search_item_type() with material types"""
+        for material_type in GENERIC_MATERIAL_TYPES:
+            item = {"type": material_type}
```
mikkonie: More double quotes here and below in the same file.

samplesheets/tests/test_templatetags.py
```diff
@@ -0,0 +1,227 @@
...
+        self.assertIn("text-info", icon_html)
+        self.assertIn("mdi:folder-table", icon_html)
+
+    def test_get_icon_assay(self):
+        """Test get_icon() with Assay"""
+        icon_html = s_tags.get_icon(self.assay)
+        self.assertIn("text-danger", icon_html)
+        self.assertIn("mdi:table-large", icon_html)
+
+    def test_get_isatab_tag_html_with_tags_in_TAG_COLORS(self):
```
mikkonie: Stlye note, let's not have capitalized letters in method names.

samplesheets/tests/test_templatetags.py
```diff
@@ -0,0 +1,227 @@
...
+        irods_request = type(
+            "MockIrodsRequest", (object,), {"status": "UNKNOWN"}
+        )
+        self.assertEqual(s_tags.get_request_status_class(irods_request), "")
+
+    def test_trim_base_path(self):
+        """Test trim_base_path() with a realistic iRODS path"""
+        prefix = "/base_path"
+        path = prefix + "/project/subfolder1/subfolder2"
+        expected_trimmed_path = "/project/subfolder1/subfolder2"
```
mikkonie: For consistency and clarity, we should call variables like this in tests just `expected`, unless several different variables are needed within one test case.

samplesheets/tests/test_views.py
```diff
@@ -691,7 +691,7 @@ def test_render_with_sheets(self):
             )
 
 
-class TestSheetTemplateCreateFormView(TestViewsBase):
+class TestSheetTemplateCreateView(TestViewsBase):
     """Tests for SheetTemplateCreateFormView"""
```
mikkonie: The class name was updated but the old name remains in the class docstring :)

mikkonie (changes requested):
Looks good and works just as expected, good work. I have some requests for changes, mainly related to style issues.

These should preferably have been two separate PRs as they are not related in any other way than belonging in the same app, but it's not a big deal since it's not a huge PR.

samplesheets/tests/test_templatetags.py
```diff
@@ -0,0 +1,225 @@
...
+
+    def test_get_irods_tree(self):
+        """Test get_irods_tree()"""
+        ret = s_tags.get_irods_tree(self.investigation)
+        # Assert that IRODS_SAMPLE_COLL exists in the returned string
+        self.assertIn(IRODS_SAMPLE_COLL, ret)
+        # Assert that study path exists in the returned string
+        self.assertIn(self.irods_backend.get_sub_path(self.study), ret)
+        # Assert that assay path exists in the returned string
+        clean_assay_path_0 = self.irods_backend.get_sub_path(self.assay).split(
```
mikkonie: It seems you are needlessly splitting the same path twice here.

Also, minor notes on conventions: We don't traditionally name variables in the format of `x_0`. The underscore is not needed. Also, since these are only internal variables used inside this one test case, feel free to shorten them a bit :)

samplesheets/tests/test_templatetags.py
```diff
@@ -0,0 +1,225 @@
...
+            kwargs={'project': self.project.sodar_uuid},
+        )
+        expected_url += '#/study/{}/filter/Sample1'.format(
+            self.study.sodar_uuid
+        )
+        self.assertEqual(url, expected_url)
+
+    def test_get_irods_path_with_project(self):
+        """Test get_irods_path() with project"""
+        expected_path = self.irods_backend.get_path(self.project)
```
mikkonie: We can just name these variables `expected`, unless there are multiple expected values within a single test cases we need to differentiate between.

mikkonie (changes requested):
A couple of minor notes, can merge after these have been handled.

samplesheets/tests/test_templatetags.py
```diff
@@ -0,0 +1,216 @@
...
+
+    def test_get_irods_tree(self):
+        """Test get_irods_tree()"""
+        ret = s_tags.get_irods_tree(self.investigation)
+        # Assert that IRODS_SAMPLE_COLL exists in the returned string
+        self.assertIn(IRODS_SAMPLE_COLL, ret)
+        # Assert that study path exists in the returned string
+        self.assertIn(self.irods_backend.get_sub_path(self.study), ret)
+        # Assert that assay path exists in the returned string
+        path0, path1 = self.irods_backend.get_sub_path(self.assay).split('/')
```
mikkonie: I'm not sure if I understand what happens here. You split the same path into two parts and then assert both of them are found in the return data. Why not just assert the inclusion of the full path?

mikkonie (changes requested):
Reviewed, had one comment.

samplesheets/tests/test_templatetags.py
gromdimon: The reason for that is response. It is html, which is formed with tags:
`<ul><li>sample_data<ul><li>study_dff42df4-362a-460c-86cf-749897eb6190<ul><li>assay_f9fb2c7d-9541-4a44-a8a3-0bbaebfdb7ce</li></ul></li></ul></li></ul>`
So I've splited path and asserted it separately, as it was the simplest solution :)

mikkonie (approved):
Looks good, will merge.

PR #1729, opened 2023-07-13
Create and use constants for landing zone status types

### LINKED ISSUE #1398: Create and use constants for landing zone status types
Something I noticed when doing #691. `ZONE_STATUS_TYPES` and all the other variables simply use strings, I generally try avoid that because typos make for nasty bugs.

This is a minor thing, but since we are updating landing zone UI features, might as well do this one as well!

## Spec

- Specify constants for zone status types in `landingzones.models`
    * E.g. `'OK'` -> `ZONE_STATUS_OK`
- Refactor usages of status types to use these constants
- Ensure tests pass :)

**NOTE:** It's probably better to do this **after** we merge the v0.13 updates, might create a lot of merge conflicts otherwise..

gromdimon:
### TODO:
- [x] Implement
- [x] Test

### DIFF
--- a/irodsadmin/management/commands/irodsorphans.py
+++ b/irodsadmin/management/commands/irodsorphans.py
@@ -12,6 +12,7 @@ from projectroles.models import Project
 from projectroles.plugins import get_backend_api
 
 # Landingzones dependency
+from landingzones.constants import ZONE_STATUS_MOVED, ZONE_STATUS_DELETED
 from landingzones.models import LandingZone
 
 # Samplesheets dependency
@@ -84,7 +85,7 @@ def get_zone_collections(irods_backend):
     return [
         irods_backend.get_path(lz)
         for lz in LandingZone.objects.filter(
-            ~(Q(status='MOVED') & Q(status='DELETED'))
+            ~(Q(status=ZONE_STATUS_MOVED) & Q(status=ZONE_STATUS_DELETED))
         )
     ]
 
--- a/landingzones/constants.py
+++ b/landingzones/constants.py
@@ -1,0 +1,91 @@
+"""Constants for the landingzones app"""
+
+# Status types for landing zones
+ZONE_STATUS_OK = 'OK'
+ZONE_STATUS_CREATING = 'CREATING'
+ZONE_STATUS_NOT_CREATED = 'NOT CREATED'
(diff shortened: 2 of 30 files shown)


mikkonie (changes requested):
Looking good! I can see two issues:

## Issue 1

Importing large amounts of these constants in modules can bloat up the import sections quite a bit. I suggest you move these and the other landingzones constants into `landingzones/constants.py`. Then, if you need a big bunch of these constants imported in a single module, you can instead do:

```
import landingzones.constants as lc
do_something_with(lc.ZONE_STATUS_OK)  # Etc etc.
```

## Issue 2

Grep shows me there are still places where we refer to these status types by string instead of the constant. Mostly in tests but also elsewhere. Example:

```
(sodar14) mikkopen@cubi-7300-mn:~/code/sodar$ grep -niIr --include "*.py" "'MOVED'"
samplesheets/views.py:1358:                    status__in=['MOVED', 'DELETED']
landingzones/management/commands/inactivezones.py:23:    ).exclude(status__in=('DELETED', 'MOVED'))
landingzones/plugins.py:128:        if obj.__class__ == LandingZone and obj.status != 'MOVED':
landingzones/plugins.py:169:        active_count = zones.exclude(status__in=['MOVED', 'DELETED']).count()
landingzones/models.py:24:ZONE_STATUS_MOVED = 'MOVED'
landingzones/models.py:75:STATUS_FINISHED = ['MOVED', 'NOT CREATED', 'DELETED']
irodsadmin/management/commands/irodsorphans.py:87:            ~(Q(status='MOVED') & Q(status='DELETED'))
taskflowbackend/tests/test_flows.py:632:        self.assertEqual(self.zone.status, 'MOVED')
taskflowbackend/tests/test_flows.py:691:        self.assertEqual(self.zone.status, 'MOVED')
taskflowbackend/tests/test_flows.py:763:        self.assertEqual(self.zone.status, 'MOVED')
taskflowbackend/tests/test_flows.py:967:        self.assertEqual(new_zone.status, 'MOVED')
taskflowbackend/flows/landing_zone_move.py:329:                    'status': 'MOVED',
```

We should consistently use the constants everywhere.

mikkonie (approved):
No notes, good work! Will be merged :)

PR #1754, opened 2023-08-18
Crash in `ZoneStatusRetrieveAjaxView` views from `getlist()`

### LINKED ISSUE #1752: Crash in ZoneStatusRetrieveAjaxView views from getlist()
In staging we observed a crash in `ZoneStatusRetrieveAjaxView`, dump below in comments.

Looks like we're trying to use `getlist()` on a dict, which is generally not a good idea :)

**How to Reproduce:** Navigate to `ProjectZonesView` in a project with no landing zones available.

## Spec

- Fix the landingzones JQuery to not call for the Ajax view if no zones are in the UI
- Fix the Ajax view to gracefully handle not receiving any zones in its call
    * Just in case it does get called
    * Just return an empty status_dict and a 200
- Add an ajax view to test for the view handling an "empty" request

gromdimon:
### Tasks:
- [x] Fix the bug
- [x] Test it

### DIFF
--- a/landingzones/static/landingzones/js/landingzones.js
+++ b/landingzones/static/landingzones/js/landingzones.js
@@ -20,7 +20,7 @@ var updateZoneStatus = function() {
     });
 
     // Make the POST request to retrieve zone statuses
-    if (zoneUuids.length >= 0) {
+    if (zoneUuids.length > 0) {
         $.ajax({
             url: zoneStatusUrl,
             method: 'POST',
--- a/landingzones/static/landingzones/js/landingzones.js
+++ b/landingzones/static/landingzones/js/landingzones.js
@@ -6,7 +6,7 @@ var isSuperuser = false;
  *****************************/
 var updateZoneStatus = function() {
     window.zoneStatusUpdated = false;
-    var zoneUuids = '';
+    var zoneUuids = [];
 
     $('.sodar-lz-zone-tr-existing').each(function() {
         var trId = $(this).attr('id');
@@ -15,7 +15,7 @@ var updateZoneStatus = function() {
         var statusTd = zoneTr.find('td#sodar-lz-zone-status-' + zoneUuid);
 
         if (statusTd.text() !== 'MOVED' && statusTd.text() !== 'DELETED') {
-            zoneUuids += zoneUuid + ',';
+            zoneUuids.push(zoneUuid);
         }
     });
 
--- a/landingzones/static/landingzones/js/landingzones.js
+++ b/landingzones/static/landingzones/js/landingzones.js
@@ -6,117 +6,118 @@ var isSuperuser = false;
  *****************************/
 var updateZoneStatus = function() {
     window.zoneStatusUpdated = false;
-    var zoneUuids = [];
+    var zoneUuids = '';
 
     $('.sodar-lz-zone-tr-existing').each(function() {
         var trId = $(this).attr('id');
         var zoneTr = $('#' + trId);
         var zoneUuid = $(this).attr('data-zone-uuid');
-        var sampleUrl = $(this).attr('data-sample-url');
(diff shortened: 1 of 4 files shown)


landingzones/static/landingzones/js/landingzones.js
```diff
@@ -6,117 +6,118 @@ var isSuperuser = false;
...
-
-                if (
-                    statusTd.text() !== zoneStatus ||
-                    statusInfoSpan.text() !== zoneStatusInfo
-                ) {
-                    statusTd.text(zoneStatus);
-                    statusTd.removeClass();
-                    statusTd.addClass(statusStyles[zoneStatus] + ' text-white');
-                    statusInfoSpan.text(zoneStatusInfo);
+    if (zoneUuids.length >= 0) {
```
mikkonie: I'm still seeing requests to the Ajax view even if there are no zones for which to check the status. The condition here should probably be `zoneUuids.length > 0`.

landingzones/tests/test_views_ajax.py
```diff
@@ -26,3 +26,16 @@ def test_post(self):
             }
         }
         self.assertEquals(response.data, expected)
+
+    def test_post_no_landingzone(self):
```
mikkonie: Minor style point, `test_post_no_zone()` would probably be more consistent with the naming of other tests. Otherwise this looks good!

mikkonie (changes requested):
Reviewed, found one bug and a style note.

landingzones/static/landingzones/js/landingzones.js
gromdimon: Forgot to return to the `>` after testing :)

landingzones/static/landingzones/js/landingzones.js
```diff
@@ -6,117 +6,118 @@ var isSuperuser = false;
...
     $('.sodar-lz-zone-tr-existing').each(function() {
         var trId = $(this).attr('id');
         var zoneTr = $('#' + trId);
         var zoneUuid = $(this).attr('data-zone-uuid');
-        var sampleUrl = $(this).attr('data-sample-url');
         var statusTd = zoneTr.find('td#sodar-lz-zone-status-' + zoneUuid);
 
         if (statusTd.text() !== 'MOVED' && statusTd.text() !== 'DELETED') {
-            zoneUuids.push(zoneUuid);
+            zoneUuids += zoneUuid + ',';
```
mikkonie: This latest update to the pr works now, and that's great. However, we should do this in a more standardized way.

The original way of sending this as a list is fine. The problems were that:
1. We were performing the request even when there was nothing to queryt for, and
2. The Ajax view didn't correctly handle not receiving a list.

Converting back and forth between strings adds extra complexity which we don't actually require.

landingzones/views_ajax.py
```diff
@@ -22,7 +22,8 @@ def check_zone_permission(self, zone, user):
         return user.has_perm(permission, obj=zone.project)
 
     def post(self, request, *args, **kwargs):
-        zone_uuids = request.data.getlist('zone_uuids[]')
+        zone_uuids_string = request.data.get('zone_uuids')
+        zone_uuids = zone_uuids_string.split(',')[:-1]
```
mikkonie: As commented in the javascript file, we should still provide and grab a standard list like in the original implementation. Here in the view we just need to check if the value we're looking at is an instance of `QueryDict` before we call `getlist()`.

mikkonie (changes requested):
This works now, good job! However, I would like one last bit of refactoring.

mikkonie (approved):
This still works as expected, which is great. The check in the ajax view I specifically asked for is missing, however it doesn't seem that is required with the current way we make requests to the Ajax view. Will merge.

PR #1915, opened 2024-02-28
Added settings needed for reverse proxy setup

HaraldWilhelmi:
We need to put our SODAR installation behind another reverse proxy.
For this purpose we exposed some more Django settings via environment variables.
A word of warning about the general topic:

* The present, official Django documentation states that CSRF_TRUSTED_ORIGINS should be a list of **URL patterns** (e.g. 'https://*.example.com').
* The Django version presently used in our SODAR container (3.2.24) uses **domain names** or **domain name suffixes** (e.g.  '.example.com' or 'sodar.example.com'). That I learned by reading the source.

So a upgrade of the Django version used may change that.


config/settings/base.py
```diff
@@ -35,6 +35,7 @@
 # ------------------------------------------------------------------------------
 # Hosts/domain names that are valid for this site
 ALLOWED_HOSTS = env.list('DJANGO_ALLOWED_HOSTS', default=['*'])
+USE_X_FORWARDED_HOST = env.bool('DJANGO_USE_X_FORWARDED_HOST', default=False)
```
mikkonie, line 38: For `env.bool` and `env.str`, we don't actually need the `default=` kwarg, just using an arg is enough. For `env.list` this is needed, unless `cast` is also defined as an arg.

Some existing values in the SODAR config have the unnecessary `default` kwargs, I believe these came from the original Cookiecutter template. Those will be cleaned up eventually as well :)

mikkonie (changes requested):
Looking good except for one minor cosmetic note.

Please open a ticket for this PR in the issue tracker. That will make it easier to follow up on and refer to in the future. Other than that, I'm happy to merge.

The next major SODAR version (v1.0) will be upgraded to Django 4.2 so, indeed, we'll have to see what changes it will require. This work is already underway in the [SODAR Core repository](https://github.com/bihealth/sodar-core).

HaraldWilhelmi:
Thanks for the feedback. Will re-file a pull request with changes and a ticket.

config/settings/base.py
HaraldWilhelmi: I'm afraid we actually need "default=False" in the boolean case.
If I remove it, building the Docker image will fail like this:

`django.core.exceptions.ImproperlyConfigured: Set the DJANGO_USE_X_FORWARDED_HOST environment variable`

Issue #1917, opened 2024-02-29
Allow setup behind reverse proxy


HaraldWilhelmi:
### Problem

If SODAR is set up with docker-compose and an extra reverse proxy is put in front of it (e.g. as part of a firewall), the login will fail due to CSRF checks inside Django.

### Solution

To make such setup work, SODAR should expose some more Django settings, related to the CSRF checks:

* CSRF_TRUSTED_ORIGINS
* CSRF_COOKIE_DOMAIN
* USE_X_FORWARDED_HOST

### Additional Context

The precise meanings of the above settings depend on the Django version. That is especially true for CSRF_TRUSTED_ORIGINS. Currently SODAR uses version, which requires this parameter to hold a list of DNS names and DNS name suffixes as documented here:

* https://docs.djangoproject.com/en/3.2/ref/settings/#csrf-trusted-origins

Future versions will most likely require URL patterns instead, as documented in the most recent Django documentation.

Will file a pull request soon.

mikkonie:
Done by @HaraldWilhelmi and merged.

PR #1918, opened 2024-02-29
 Allow setup behind reverse proxy #1917 

HaraldWilhelmi:
That is the pull request for feature #1917. Seems that we need to stick with the default for USE_X_FORWARDED_HOST because otherwise the Docker build fails.


config/settings/base.py
```diff
@@ -35,6 +35,7 @@
 # ------------------------------------------------------------------------------
 # Hosts/domain names that are valid for this site
 ALLOWED_HOSTS = env.list('DJANGO_ALLOWED_HOSTS', default=['*'])
+USE_X_FORWARDED_HOST = env.bool('DJANGO_USE_X_FORWARDED_HOST', default=False)
```
mikkonie: Please remove the redundant `default` kwarg and just provide `False`, same with `env.str()` in `production.py`. Otherwise looking good.

mikkonie (changes requested):

config/settings/base.py
HaraldWilhelmi: I'm afraid the 'default' is required. Without the default I can build the Docker image:

`django.core.exceptions.ImproperlyConfigured: Set the DJANGO_USE_X_FORWARDED_HOST environment variable`

mikkonie: In the same settings file, there are multiple examples of `env.bool()` working with the default provided as an arg instead of a kwarg. This is just a variable like any other so it should work as well. Maybe you've made some other mistake?

HaraldWilhelmi: I believe that this works in the other case just due to some (?) default setting in the Docker setup for that variable.
IMHO the default setting is not redundant, as one can see from this small experiment:

```
$ unset FOO
$ python
Python 3.10.11 (main, May 15 2023, 15:04:04) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import environ
>>> env = environ.Env()
>>> print(env.bool('FOO', default=False))
False
>>> print(env.bool('FOO'))
Traceback (most recent call last):
  File "/home/wilhelmh/.virtualenvs/sodar-server/lib/python3.10/site-packages/environ/environ.py", line 403, in get_value
    value = self.ENVIRON[var_name]
  File "/usr/local/python3.10/lib/python3.10/os.py", line 680, in __getitem__
    raise KeyError(key) from None
KeyError: 'FOO'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/wilhelmh/.virtualenvs/sodar-server/lib/python3.10/site-packages/environ/environ.py", line 246, in bool
    return self.get_value(var, cast=bool, default=default)
  File "/home/wilhelmh/.virtualenvs/sodar-server/lib/python3.10/site-packages/environ/environ.py", line 407, in get_value
    raise ImproperlyConfigured(error_msg) from exc
django.core.exceptions.ImproperlyConfigured: Set the FOO environment variable
```

In case you want to drop the default in the code, we have to add it elsewhere.

mikkonie: The default should of course not be dropped. It should be provided as an arg, not a kwarg.

HaraldWilhelmi: Sorry - completely missed your point. Will fix that.

HaraldWilhelmi: Have just submitted another squashed commit.

mikkonie: I could have been more clear, apologies. Looking good now, I'll merge :)

mikkonie (approved):

PR #1945, opened 2024-06-12
feat: Add generic assayapp featuring dynamic links

### LINKED ISSUE #1868: Expand inline linking of ms_meta assayapp
Our MS Metabolomics ISA model is currently being augmented in collaboration with BIH Metabolomics Core to accommodate more metadata features. SODAR's ms_meta assayapp should be expanded to create more inline links for some of these features.

New ISA model (WIP):
ms_meta

### LINKED ISSUE #1946: Add generic assay plugin
### Problem

Currently linking of the samplesheets metadata to the iRODS data collections is facilitated by assay plugins (assayapps) which perform a pre-defined set of tasks depending on the type of assays we're expecting.

https://sodar-server.readthedocs.io/en/dev/metadata_advanced.html

This means, for every new assay we want to support, a new assay plugin needs to be introduced to the SODAR codebase.  

### Solution

We want to create a new "universal" plugin which takes information from special comments in the `i_Investigation.txt` to define a data linking scheme. This is more flexible than pre-defining one scheme per assay and in-line with the existing "SODAR Assay Plugin" special comment which can be used to overwrite automatic plugin selection.

The logic introduced could potentially be integrated into the main SODAR code and GUI at a later time.

sellth:
This is a test balloon for implementing a new dynamic way of defining SODAR data links. This assayapp is only active when specifically called via the `Comment[SODAR Assay Plugin]` line in *i_Investigation.txt*. Further special comments define columns which should link to different collections.

- [x] Comment[SODAR Assay Plugin ResRep]: Links columns to `Results & Reports`
- [x] Comment[SODAR Assay Plugin MiscFiles]: Links columns to `MiscFiles`
- [x] Comment[SODAR Assay Plugin Data1/2/3]: Creates row links to a three-tiered folder structure in iRODS.
- [x] Comment[SODAR Assay Plugin RowPath]: Create links to data collections defined by previous comment.
- [x] Toggle display row links via comment (eed609e76da465ea657d7ed859a3343702667365)
- [x] docs

Involving @shubavarshini and @mikkonie right from the start to get feedback.


sellth:
Added row-link generation for three-level data collection structure. Do we need to also care about Material columns i.e. top_header shenanigans?

samplesheets/assayapps/generic/plugins.py
```diff
@@ -0,0 +1,175 @@
...
+from samplesheets.rendering import SIMPLE_LINK_TEMPLATE
+from samplesheets.utils import get_top_header
+from samplesheets.views import MISC_FILES_COLL, RESULTS_COLL
+
+
+# Local constants
+APP_NAME = 'samplesheets.assayapps.generic'
+RESULTS_REPORTS_COMMENT = 'SODAR Assay Plugin ResRep'
+MISC_FILES_COMMENT = 'SODAR Assay Plugin MiscFiles'
+DATA_COMMENTS = [
```
mikkonie: We probably don't want to use a fixed amount of hardcoded values here. Maybe this variable could define the prefix (e.g. `SODAR Assay Plugin Data`) and then we could iterate for as long the prefix + a running integer is found in the comments.

samplesheets/assayapps/generic/plugins.py
```diff
@@ -0,0 +1,175 @@
...
+    'SODAR Assay Plugin Data1',
+    'SODAR Assay Plugin Data2',
+    'SODAR Assay Plugin Data3',
+]
+
+
+class SampleSheetAssayPlugin(SampleSheetAssayPluginPoint):
+    """Plugin for generic assays in sample sheets"""
```
mikkonie: Minor note, but this should probably say "generic data linking", as "generic assay" sounds like a very specific concept :)

samplesheets/assayapps/generic/plugins.py
```diff
@@ -0,0 +1,175 @@
...
+                    .split(';')
+                )
+                if header['value'].lower() in ResRepCols:
+                    row[i]['value'] = SIMPLE_LINK_TEMPLATE.format(
+                        label=row[i]['value'],
+                        url=base_url
+                        + '/'
+                        + RESULTS_COLL
+                        + '/'
+                        + row[i]['value'],
```
mikkonie: For these long path joins it might look cleaner to use `os.path.join()`. And before you say it, yes, I probably have to clean things up similarly in my own code :)

samplesheets/assayapps/generic/plugins.py
```diff
@@ -0,0 +1,175 @@
...
+
+    #: Required permission for accessing the plugin
+    # TODO: TBD: Do we need this?
+    permission = None
+
+    #: Toggle displaying of row-based iRODS links in the assay table
+    display_row_links = True
+
+    @classmethod
+    def _get_col_value(cls, colname, row, table):
```
mikkonie: `colname` should be named `col_name` for consistency.

samplesheets/assayapps/generic/plugins.py
```diff
@@ -0,0 +1,175 @@
...
+
+        for i in range(len(row)):
+            header = table['field_header'][i]
+            if not top_header or i >= th_colspan:
+                top_header = get_top_header(table, i)
+                th_colspan += top_header['colspan']
+
+            # TODO: Check if two comments reference the same column header?
+            # Create Results & Reports links
+            if assay.comments.get(RESULTS_REPORTS_COMMENT):
```
mikkonie: There seems to be repeated code for the results and reports collections and the misc files collections. Maybe make a single method out of that and call it with the comment as an arg.

mikkonie (changes requested):
Looking good, the I had a few minor comments. I didn't test this out in practice yet but I will :)

samplesheets/assayapps/generic/plugins.py
sellth: good idea

samplesheets/assayapps/generic/plugins.py
sellth: Maybe an fstring is even cleaner as it does not require importing os?

samplesheets/assayapps/generic/plugins.py
sellth: implemented

samplesheets/assayapps/generic/plugins.py
sellth: implemented, it's a bit messy though

samplesheets/assayapps/generic/plugins.py
```diff
@@ -0,0 +1,186 @@
...
+            data_collections.append(
+                self._get_col_value(column_name, row, table)
+            )
+
+        # Build iRODS path from list and stop at first None value
+        data_path = ''
+        for col in data_collections:
+            if col:
+                data_path += f'/{col}'
+            else:
```
mikkonie: We should only append values into `data_collections` while they are not `None`. Then, we can simply use `os.path.join()` here to join the columns  with actual values together. Inserting `None` into the list is redundant.

samplesheets/assayapps/generic/plugins.py
```diff
@@ -0,0 +1,186 @@
...
+        """
+        Creates collection links for targeted columns.
+
+        :param cell: Dict (obtained by iterating over a row)
+        :param header: Column header
+        :param top_header: Column top header
+        :param target_cols: List of column names.
+        :param url: Base URL for link target.
+        """
+
```
mikkonie: Blank line at the start of the method, should be removed.

samplesheets/assayapps/generic/plugins.py
```diff
@@ -0,0 +1,186 @@
...
+            return row
+        assay_path = self.get_assay_path(assay)
+        if not assay_path:
+            return row
+
+        base_url = settings.IRODS_WEBDAV_URL + assay_path
+        top_header = None
+        th_colspan = 0
+
+        results_cols = assay.comments.get(RESULTS_COMMENT)
```
mikkonie, line 152: Question about the overall logic: Why do we only allow these auto-generated inline links to point to `ResultsReports` and `MiscFiles` collections? Why not within the actual data collection associated with the row?

The general idea of data linking in SODAR is exactly the reverse: it is generally assumed that assay row specific files go into the row data collection and that's where the assay data points at. `ResultsReports` and `MiscFiles` are intended to contain more general files, not necessarily associated with e.g. a single sample.

It's of course ok to extend this so that we can also link in those general collections from the row cells, but why limit it to that? It seems to be going directly against the current practices. Hence this will not be a "generic" plugin but quite a specialized one.

Or, is it assumed that inline links to the row data collection have already been "hardcoded" into the assay table as simple links and thus we don't have to update them on the fly with this update method?

Maybe we discussed this before, but in that case I have forgotten.

samplesheets/assayapps/generic/plugins.py
```diff
@@ -0,0 +1,186 @@
...
+            # Create MiscFiles links
+            if misc_cols:
+                self._link_from_comment(
+                    row[i],
+                    header,
+                    top_header,
+                    misc_cols,
+                    f'{base_url}/{MISC_FILES_COLL}',
+                )
+
```
mikkonie: Unnecessary blank line here.

samplesheets/assayapps/generic/plugins.py
```diff
@@ -0,0 +1,186 @@
...
+        """
+        Return value of last matched column.
+
+        :param target_col: Column name to look for
+        :param row: List of dicts (a row returned by SampleSheetTableBuilder)
+        :param table: Full table with headers (dict returned by
+                      SampleSheetTableBuilder)
+        :return: String with cell value of last matched column
+        """
+        # returns last match of row
```
mikkonie: This should start capitalized.

samplesheets/assayapps/generic/plugins.py
```diff
@@ -0,0 +1,186 @@
...
+        :param target_cols: List of column names.
+        :param url: Base URL for link target.
+        """
+
+        if header['value'].lower() in target_cols:
+            cell['value'] = SIMPLE_LINK_TEMPLATE.format(
+                label=cell['value'],
+                url=f"{url}/{cell['value']}",
+            )
+        elif top_header['value'].lower() in target_cols:
```
mikkonie: Using the top header here and allowing any top header seems ill-advised. There's usually even more repetition here than on the header level. Also, with materials/processes containing more than one column, this will apply the same linking logic to each cell within that node. Maybe there are cases in which this is what the user wants, but how common is it?

I guess this works for data file columns, but those links generally should point to data collections and not ResultsReports/MiscFiles anyway..

This feels like an easy way for the user to shoot themselves in the foot and generate a lot of unwanted (and probably not working) links. Or is there a clear use case for this? I may just not be seeing it.

Perhaps this could be limited to certain top header names where this actually makes sense.

samplesheets/assayapps/generic/plugins.py
```diff
@@ -0,0 +1,186 @@
...
+        Creates collection links for targeted columns.
+
+        :param cell: Dict (obtained by iterating over a row)
+        :param header: Column header
+        :param top_header: Column top header
+        :param target_cols: List of column names.
+        :param url: Base URL for link target.
+        """
+
+        if header['value'].lower() in target_cols:
```
mikkonie, line 74: Going by a single header name is, of course, risky. The same header names usually get repeated a lot within sample sheets, so this gives the user power to generate a lot of non-working links.

samplesheets/assayapps/generic/plugins.py
```diff
@@ -0,0 +1,186 @@
...
+            if col:
+                data_path += f'/{col}'
+            else:
+                break
+
+        if data_path:
+            return assay_path + data_path
+        return None
+
+    def update_row(self, row, table, assay):
```
mikkonie: There are some problems if intending to do a lot of the file linking using `update_row()`:

- At least for now, the results for this method are not cached. This means the code will be run for each assay row on each page load in non-edit mode, if iRODS collections have been created. This will slow things up with large sheets. Guess why we have the caching to begin with :)
    * Adding caching of updated row values if of course possible, and practically mandatory if we intend to use this way of linking more, but that's a different task and not exactly trivial..
- Your implementation probably should also take into account if e.g. a cell already contains a simple link. If not checked, you'll just generate an invalid URL.

samplesheets/assayapps/generic/plugins.py
```diff
@@ -0,0 +1,186 @@
...
+        :param assay: Assay object
+        :return: List of dicts
+        """
+        if not settings.IRODS_WEBDAV_ENABLED or not assay:
+            return row
+        assay_path = self.get_assay_path(assay)
+        if not assay_path:
+            return row
+
+        base_url = settings.IRODS_WEBDAV_URL + assay_path
```
mikkonie, line 148: You probably want to move everything up to `base_url = # etc` into the class constructor. This code will be repeated for each assay row, which will not perform well on large tables. Especially since `get_assay_path()` requires a Django database query (which I will refactor away now that I've noticed that, but nevertheless)

This hasn't really been a problem with current use of `update_row()`, which is mostly for special cases. If we intend to use it for other things, we might try to save all the cycles we can on per-row operations.

mikkonie (commented):
I looked deeper into this and found some questions about the overall logic, here in this review.

Also one small request: in the future please use the branch naming conventions of the project, namely `issue_id-short-description`, as it makes easier to follow up on work branches.

samplesheets/assayapps/generic/plugins.py
sellth: I would claim this is a feature. All columns of the same name (i.e. "Report File" comments in various processes) will link to files in the same location.

samplesheets/assayapps/generic/plugins.py
sellth: This part is indeed for creating links of "Data File" materials, as this column name is actually in the top_header and not the header. So we could very well create a whitelist for this and only create links for header == "Name" columns. This method is creating arbitrary links btw, not only to MiscFiles and ResultsReports.

Side note: The SIMPLE_LINK_TEMPLATE is not working for Data File columns, correct?

samplesheets/assayapps/generic/plugins.py
sellth: Makes sense. I'm still hesitant to use os.path.join() though since it's not a real path.

samplesheets/assayapps/generic/plugins.py
sellth: Limited to DATA_FILE_HEADERS and MATERIAL_NAME_HEADERS

samplesheets/assayapps/generic/plugins.py
sellth: I now check if cell['value'] is str and if it matches this regex: '.+ <.*>'

shubavarshini (dismissed):
All good. I understood the code but couldn't envision it in the ISA tabs. Maybe I'm forgetting/confusing something that we have already discussed.

sellth:
@mikkonie @shubavarshini  I believe we're ready. Main changes are naming of the ISA-tab comments and docs.

sellth:
@mikkonie @shubavarshini I believe we're ready. Main changes are naming of the comments and adding docs.

docs_manual/source/metadata_advanced.rst
```diff
@@ -107,6 +108,7 @@ SODAR currently supports the following assay plugins:
 - **Microarray**
 - **Protein Expression Profiling / Mass Cytometry**
 - **Protein Expression Profiling / Mass Spectrometry**
+- **Universal Assay Plugin**
```
mikkonie: Let's call this "Generic Assay Plugin" to stay consistent with SODAR language and the internal naming.

This might initially make it confusing with the "Generic Raw Data Plugin". However, I will handle it by renaming/removing that plugin in a subsequent commit. It isn't currently used in any project in our SODAR instance and we could simply override its functionality with this plugin anyway.

docs_manual/source/metadata_advanced.rst
```diff
@@ -216,3 +218,40 @@ Protein Expression Profiling / Mass Cytometry Plugin
     * *Raw Data Files* and *Derived Data Files* are linked to ``{Assay Name}``
 - Used with measurement type / technology type
     * protein expression profiling / mass cytometry
+
+Universal Assay Plugin
```
mikkonie: Here we should also use the name "Generic Assay Plugin".

docs_manual/source/metadata_advanced.rst
```diff
@@ -216,3 +218,40 @@ Protein Expression Profiling / Mass Cytometry Plugin
...
+    * Each comment should define one column name.
+    * Values of those columns are used to define the ``RowPath``.
+    * For example:
+        .. code-block::
+
+            STUDY ASSAYS
+            Comment[SODAR Assay Row Path 1]    Pool ID
+            Comment[SODAR Assay Row Path 2]    Extract Name
+
+        + Resulting row links:
```
mikkonie: This is misaligned when the doc is rendered. I guess it should be de-indented to the previous level.

plugin-doc

docs_manual/source/metadata_advanced.rst
```diff
@@ -216,3 +218,40 @@ Protein Expression Profiling / Mass Cytometry Plugin
     * *Raw Data Files* and *Derived Data Files* are linked to ``{Assay Name}``
 - Used with measurement type / technology type
     * protein expression profiling / mass cytometry
+
+Universal Assay Plugin
+----------------------
+
+This plugin enables the user to define row-specific and inline links to
+iRODS collections via comments in the ``STUDY ASSAYS`` section of the ISA-Tab
+investigation file.
```
mikkonie: Maybe add further clarification like "This plugin can be used with any assay measurement and technology configuration, as long as the desired linking is given in ISA-Tab comments"?

By the way, this description of the plugin is a good idea. I'll have to add similar short ones for the others as well.

docs_manual/source/metadata_advanced.rst
```diff
@@ -216,3 +218,40 @@ Protein Expression Profiling / Mass Cytometry Plugin
...
+
+- Internal name: ``samplesheets_assay_generic``
+- Row-specific links
+    * Place one or multiple comments starting with ``SODAR Assay Row Path``.
```
mikkonie: We should make it clear here that the ordering of the comments matters. The user should be explicitly encouraged to number them in order.

In the code, you correctly do sorting of the dictionary keys to ensure this. However, despite the nice example below, this documentation could lead the user to think each comment name can be whatever, as long as it starts with the given string.

samplesheets/assayapps/generic/plugins.py
```diff
@@ -0,0 +1,206 @@
...
+            if results_cols:
+                self._link_from_comment(
+                    row[i],
+                    header,
+                    top_header,
+                    results_cols,
+                    f'{base_url}/{RESULTS_COLL}',
+                )
+            # Create MiscFiles links
+            if misc_cols:
```
mikkonie, line 184: Here we still have the problem that we attempt to modify the same field up to three times, even if we already did.

Also, if the user has specified multiple collection types for linking and added the same field names (accidentally?) into many collection types, we may modify the same cell value multiple times and probably malign the resulting link in the process.

My suggestion: have `_link_from_comment()` return e.g. `True` if the value was modified. If it was, add a `continue` after the fuction call and skip the rest.

In a perfect word, we should also warn the user that they have configured the plugin in a way where they attempt to link a column for multiple different collections. However, at the moment there is no callback to e.g. Django messages in these assay plugins, so the warning would only be visible in the system logs. I'll have to think about how to add a warning mechanism to these plugins in the future.

mikkonie (changes requested):
Looking good. I have requests for some documentation improvements and one implementation-related concern.

docs_manual/source/metadata_advanced.rst
sellth: This was intentional as it is part of the example. You are not a fan of the formatting?

samplesheets/assayapps/generic/plugins.py
sellth: done

samplesheets/assayapps/generic/plugins.py
```diff
@@ -0,0 +1,210 @@
...
+            top_header['value']
+            in th.DATA_FILE_HEADERS + th.MATERIAL_NAME_HEADERS
+        ) and (header['value'] == 'Name'):
+            cell['link'] = f"{url}/{cell['value']}"
+        elif header['value'].lower() in target_cols:
+            cell['value'] = SIMPLE_LINK_TEMPLATE.format(
+                label=cell['value'],
+                url=f"{url}/{cell['value']}",
+            )
+        return True
```
mikkonie: There seems to be a bug in the implementation, as we now return `True` every time the function is called. Instead, you should have `return True` inside each of the `if`/`elif` clauses. Also, the clauses can all be changed to `if`, as we never reach the `elif` anyway once we return from inside that condition.

samplesheets/assayapps/generic/plugins.py
```diff
@@ -0,0 +1,210 @@
...
+            # Create DataCollection links
+            if data_cols:
+                if self._link_from_comment(
+                    row[i],
+                    header,
+                    top_header,
+                    data_cols,
+                    f'{settings.IRODS_WEBDAV_URL}{row_path}',
+                ):
+                    continue
```
mikkonie: The last `if` clause and `continue` can be omitted, as they are at the end of the `for` loop and we go back to the beginning here anyway.

mikkonie (changes requested):

samplesheets/assayapps/generic/plugins.py
sellth: True, but I wanted to keep the three blocks similar.

mikkonie (approved):
Looking good, thanks for the last minute changes. Will merge :)

PR #1983, opened 2024-08-15
fix: Generic assay plugin bugs

### LINKED ISSUE #1982: Generic assay plugin: Inline links in material columns always point to ResultsReports
In this example:
```
Comment[SODAR Assay Link Results]	Standardized Analysis Report
Comment[SODAR Assay Link Row]	Raw Data File;Report File
```

`Raw Data File` columns will link to `Results & Reports` when they should be pointing at the `RowPath`. This behaviour is consistent for all Data and Material headers.

### LINKED ISSUE #1984: Generic assay plugin: Samplesheets cache update fails if RowPath built from ontology column
I want to use the generic assay plugin with a new empty project. ISA-tab upload works fine, the samplesheet table is displayed. After the iRODS collections are created however the samplesheets app does not load anymore and this is in the logs:

```
sodar-celeryd-default_1  | 2024-08-30 14:58:46,807 [ERROR] samplesheets.tasks_celery: Cache update failed for project "Crashtest" (0505fa4a-efdf-4650-97ca-0498042ff64e): sequence item 1: expected str instance, list found
```

sellth:
closes #1982

### DIFF
--- a/samplesheets/assayapps/generic/plugins.py
+++ b/samplesheets/assayapps/generic/plugins.py
@@ -65,9 +65,13 @@ class SampleSheetAssayPlugin(SampleSheetAssayPluginPoint):
             return True
         # Special case for Material Names
         if (
-            top_header['value']
-            in th.DATA_FILE_HEADERS + th.MATERIAL_NAME_HEADERS
-        ) and (header['value'] == 'Name'):
+            (
+                top_header['value']
+                in th.DATA_FILE_HEADERS + th.MATERIAL_NAME_HEADERS
+            )
+            and top_header['value'].lower() in target_cols
+            and (header['value'] == 'Name')
+        ):
             cell['link'] = f"{url}/{cell['value']}"
             return True
         # Handle everything else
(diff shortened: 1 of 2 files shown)


sellth:
Now also closes #1984. What happens if we try to create links in an ontology or list column? Should we check for that?

samplesheets/assayapps/generic/plugins.py
```diff
@@ -96,7 +100,13 @@ def _get_col_value(cls, target_col, row, table):
                 header = table['field_header'][i]
                 if header['value'].lower() == target_col.lower():
                     value = row[i]['value']
-        return value
+
+        if isinstance(value, str):
+            return value
+        elif isinstance(value, list) and len(value) == 1:  # OntologyTermRefs
+            return value[0]['name']
+        else:
```
mikkonie: Here we can leave out the `else:` and simply `return None`, as reaching any of the previous clauses returns from the function anyway.

mikkonie (changes requested):
One small note, otherwise this looks good to merge.

mikkonie:
> Now also closes #1984. What happens if we try to create links in an ontology or list column? Should we check for that?

I would say these cases should be checked, as long as there's a somewhat-realistic chance someone woult attempt to create links to such columns. I'm not sure if it needs to go in this PR though or if it should be considered an additional feature. You make the call :)

sellth:
The project might end up in a state which the user cannot recover from without admin help, but it will probably be an edge case. I would prioritise getting these fixes out.

mikkonie (approved):

Issue #2375, opened 2026-01-19
Upgrade to SODAR Core v1.4


mikkonie:
Opening this issue to track stuff we should do when upgrading to SODAR Core v1.4.

## Tasks

- [ ] Upgrade Python depdendencies
- [ ] Add `martor` app in `base.py` and `urls.py`
- [ ] Add `martor` settings in `base.py` (1)
- [ ] Add `ip_address` in `AXES_LOCKOUT_PARAMETERS` default value
- [ ] Update plugin search API for v1.4 compatibility
    * bihealth/sodar-core#22
    * bihealth/sodar-core#1191
    * bihealth/sodar-core#1898
    * bihealth/sodar-core#1899
    * bihealth/sodar-core#1901
    * bihealth/sodar-core#1902
- [ ] Update search tests (2)
- [ ] Remove unused server-side search templates etc
- [ ] Use general purpose HTTP 503 exception class
    * bihealth/sodar-core#1513
- [ ] Use `HostConfirmDeleteView`
    * bihealth/sodar-core#1550

## Documentation

- [ ] Document user-facing changes to search
- [ ] Add username display setting in `ui_user_profile.rst`
- [ ] Update docs on all UI changes not mentioned here
- [ ] Add `transferroles` management command in `admin_commands.rst`
- [ ] ~~Document SODAR Core REST API version changes~~ (3)

## Notes

1. There should be no need to modify these via the env in SODAR at the moment. We can simply hardcode the default values, at least for now.
2. UI tests should not be required in this repo. Tests on `search()` output should be enough.
3. The only changed API in SODAR Core v1.4 is the projectroles sync API, which is not separately documented here. Thus, no changes should be required.

mikkonie:
Since @fmarotta has been making such awesome progress on SODAR Core v1.4.0 and there is still quite a bit of work to do on SODAR v1.3.0, I'm considering doing this upgrade already in v1.3. Let's see how this pans out!

mikkonie:
Tagging this for v1.3.0, as we seem to be nearly done with SODAR Core v1.4.0 while SODAR v1.3.0 has a bunch of work left.

mikkonie:
Done by @fmarotta with some tweaks by me, merged.

PR #2483, opened 2026-07-17
Upgrade to SODAR Core v1.4

fmarotta:
This PR addresses #2375. Most changes were in the search API, as expected, but I also encountered some other points which I'd like to discuss.

### Dependency upgrades
The `tooz` package introduced changes in recent major releases which broke the taskflow API; for the time being, I've left the package at v7.0.0. I'm happy to try and fix the problems, either here or in a separate issue, although I have no intuition about how big the task would be.

### Problems in SODAR Core
I found a bug I had previously introduced in SODAR Core (see https://github.com/bihealth/sodar-core/issues/1955 and https://github.com/bihealth/sodar-core/pull/1956), tests are failing due to this.

Another possible thing to change in Core could be the template for `HostConfirmDeleteView`. In sodar-server, the form is not shown if `can_delete_sheets` is false. This is not taken into account in the SODAR Core template, so I had to duplicate the entire template instead of extending it.


mikkonie (approved):
Looking great. I updated the docs a bit and made a few minor tweaks. Merging once CI passes.

Issue #2461, opened 2026-04-27
Add management command to find and fix orphaned IrodsAccessTicket objects


mikkonie:
Related to #2460.

The aforementioned bug has been fixed, but we may have ended up iRODS access tickets without an associated `IrodsAccessTicket` Django database object. This makes it impossible for non-superusers to manage the tickets.

We should add a management command to samplesheets which:

- Queries for iRODS access tickets
- Finds and lists access tickets with no associated `IrodsAccessTicket` object
- Recreates the deleted `IrodsAccessTicket` objects
- (Maybe in the future we can add a separate mode to delete the tickets instead, but that's for later)

## Spec

- Add management command `fixirodstickets` in `samplesheets`
    * Add `-c` / `--check` option
- Implement command logic
    * Query for all tickets on the iRODS server
        + **Note:** We don't have a "list tickets" wrapper in `irodsbackend`, at least not yet
        + See [PRC tickets documentation](https://github.com/irods/python-irodsclient/#tickets)
    * For each ticket, query for related `IrodsAccessTicket` object
        + You can determine the object by the ticket string
    * If a corresponding `IrodsAccessTicket` is not found and `-c` is not set, create the object
        + As user for the object, use the default admin
        + No reusable code for this, but that's OK, we just want to create the object
    * Output each found/recreated ticket in logs
- Add tests
    * See `samplesheets.tests.test_commands_taskflow` for examples with existing commands
        + **Note:** `test_*_taskflow` module name means we test with iRODS. It's confusing, I know :)
    * Create iRODS ticket with/without database object, assert outcome, you know the deal
    * Hint: You can use `IrodsAccessTicketMixin.make_irods_ticket()` and `IrodsAPI.issue_ticket()`
- Document command in `admin_commands.rst`

## Tasks

- [ ] Add command
- [ ] Add command logic
- [ ] Add tests
- [ ] Update docs

mikkonie:
Done by @fmarotta, merged.

PR #2491, opened 2026-07-31
Add management command to fix orphaned tickets

fmarotta:
This PR addresses #2461.

### DIFF
--- a/samplesheets/management/commands/fixirodstickets.py
+++ b/samplesheets/management/commands/fixirodstickets.py
@@ -71,6 +71,15 @@ class Command(BaseCommand):
     def _get_ticket_fields(
         cls, irods_backend: Any, ticket: dict, path: iRODSPath
     ) -> Optional[dict]:
+        """
+        Get the fields needed to construct an IrodsAccessTicket object.
+
+        :param irods_backend: IrodsAPI object
+        :param ticket: Ticket dictionary returned by the iRODS API
+        :param path: iRODSPath associated to the ticket
+        :return: Dictionary of IrodsAccessTicket fields if the ticket is valid,
+            "None" if some of the required fields are missing
+        """
         fields = {}
         ticket_string = ticket[TicketQuery.Ticket.string]
         ticket_study = irods_backend.get_uuid_from_path(path, 'study')
--- a/samplesheets/management/commands/fixirodstickets.py
+++ b/samplesheets/management/commands/fixirodstickets.py
@@ -1,0 +1,114 @@
+"""Management command to find and fix orphaned IrodsAccessTickets objects"""
+
+from datetime import datetime
+
+from irods.models import TicketQuery
+from irods.path import iRODSPath
+
+from django.conf import settings
+from django.contrib.auth import get_user_model
+from django.core.management.base import BaseCommand
+
+# Projectroles dependency
+from projectroles.management.logging import ManagementCommandLogger
+from projectroles.plugins import PluginAPI
+
+from irodsbackend.api import IrodsAPI
+
+from samplesheets.models import Assay, Study, IrodsAccessTicket
+
+
+logger = ManagementCommandLogger(__name__)
+plugin_api = PluginAPI()
+User = get_user_model()
+
+
+class Command(BaseCommand):
+    help = (
+        'Find iRODS access tickets with no associated IrodsAccessTicket object '
+        'in the database, and recreate the missing objects.'
+    )
+
+    @classmethod
+    def _find_irods_tickets(cls, irods_backend: IrodsAPI) -> tuple[str, str]:
+        """
+        Iterator for iRODS tickets.
+
+        :param irods_backend: IrodsAPI object
+        :return: Tuples of ticket query results and their associated path
+        """
+        with irods_backend.get_session() as irods:
+            coll_tickets = irods.query(
+                TicketQuery.Ticket, TicketQuery.Collection
+            )
+            data_tickets = irods.query(
+                TicketQuery.Ticket, TicketQuery.DataObject
+            )
+            for ticket in coll_tickets.get_results():
+                yield (ticket, iRODSPath(ticket[TicketQuery.Collection.name]))
+            for ticket in data_tickets.get_results():
+                yield (
+                    ticket,
+                    iRODSPath(
+                        ticket[TicketQuery.DataObject.coll],
+                        ticket[TicketQuery.DataObject.name],
+                    ),
+                )
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '-c',
+            '--check',
+            action='store_true',
+            help='Check for orphaned tickets, but do not recreate them',
+        )
+
+    def handle(self, *args, **options):
+        irods_backend = plugin_api.get_backend_api('omics_irods')
+        default_admin = User.objects.get(
+            username=settings.PROJECTROLES_DEFAULT_ADMIN
+        )
+        check = options.get('check')
+        for ticket, path in self._find_irods_tickets(irods_backend):
+            ticket_string = ticket[TicketQuery.Ticket.string]
+            try:
+                IrodsAccessTicket.objects.get(ticket=ticket_string)
+                logger.info(f'Found existing ticket object for {ticket_string}')
+            except IrodsAccessTicket.DoesNotExist:
+                ticket_study = irods_backend.get_uuid_from_path(path, 'study')
+                if ticket_study:
+                    ticket_study = Study.objects.get(sodar_uuid=ticket_study)
+                else:
+                    logger.warning(
+                        f'Ticket {ticket_string} does not have '
+                        'an associated study'
+                    )
+                    continue
+                ticket_assay = irods_backend.get_uuid_from_path(path, 'assay')
+                if ticket_assay:
+                    ticket_assay = Assay.objects.get(sodar_uuid=ticket_assay)
+                ticket_expires = ticket[TicketQuery.Ticket.expiry_ts]
+                if ticket_expires:
+                    ticket_expires = datetime.fromtimestamp(int(ticket_expires))
+                obj = IrodsAccessTicket(
+                    study=ticket_study,
+                    assay=ticket_assay,
+                    ticket=ticket_string,
+                    path=path,
+                    user=default_admin,
+                    date_created=ticket[TicketQuery.Ticket.create_time],
+                    date_expires=ticket_expires,
+                )
+                logger.info(
+                    f'Need to recreate ticket object for {ticket_string} (\n'
+                    f'\tstudy: {obj.study.sodar_uuid} ({obj.study}),\n'
+                    f'\tassay: {obj.assay.sodar_uuid if obj.assay else None},\n'
+                    f'\tticket: {obj.ticket},\n'
+                    f'\tpath: {obj.path},\n'
+                    f'\tuser: {obj.user},\n'
+                    f'\tdate_created: {obj.date_created},\n'
+                    f'\tdate_expires: {obj.date_expires},\n'
+                    ')'
+                )
+                if not check:
+                    obj.save()
--- a/samplesheets/management/commands/fixirodstickets.py
+++ b/samplesheets/management/commands/fixirodstickets.py
@@ -1,7 +1,7 @@
 """Management command to find and fix orphaned IrodsAccessTickets objects"""
 
 from datetime import datetime
-from typing import Optional
+from typing import Optional, Any
 
 from irods.models import TicketQuery
 from irods.path import iRODSPath
@@ -22,6 +22,19 @@ plugin_api = PluginAPI()
 User = get_user_model()
 
 
+# Local constants
+ASSAY_DOES_NOT_EXIST = (
+    'Ticket {ticket_string} is associated with an Assay '
+    "which doesn't exist ({ticket_assay})"
+)
+STUDY_DOES_NOT_EXIST = (
+    'Ticket {ticket_string} is associated with a Study '
+    "which doesn't exist ({ticket_study})"
+)
+STUDY_IS_MISSING = 'Ticket {ticket_string} does not have an associated study'
+TICKET_OBJECT_ACTION_COUNT = '{action} {count} orphaned ticket object{plural}'
+
+
 class Command(BaseCommand):
     help = (
         'Find iRODS access tickets with no associated IrodsAccessTicket object '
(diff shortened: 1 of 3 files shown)


samplesheets/management/commands/fixirodstickets.py
```diff
@@ -0,0 +1,114 @@
...
+
+from django.conf import settings
+from django.contrib.auth import get_user_model
+from django.core.management.base import BaseCommand
+
+# Projectroles dependency
+from projectroles.management.logging import ManagementCommandLogger
+from projectroles.plugins import PluginAPI
+
+from irodsbackend.api import IrodsAPI
```
mikkonie: We should not be importing backends directly into another apps, except in rare circumstances. Instead, we should use `PluginAPI` and `get_backend_api()`. This is already done correctly for the actual functionality. The type hint isn't very critical here, as it's an internal helper only called within this command.

samplesheets/management/commands/fixirodstickets.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+class Command(BaseCommand):
+    help = (
+        'Find iRODS access tickets with no associated IrodsAccessTicket object '
+        'in the database, and recreate the missing objects.'
+    )
+
+    @classmethod
+    def _find_irods_tickets(cls, irods_backend: IrodsAPI) -> tuple[str, str]:
```
mikkonie: Minor point, but I would name this helper simply `_get_irods_tickets()` as there isn't really any searching involed, just a simple query to get all tickets.

samplesheets/management/commands/fixirodstickets.py
```diff
@@ -0,0 +1,114 @@
...
+        parser.add_argument(
+            '-c',
+            '--check',
+            action='store_true',
+            help='Check for orphaned tickets, but do not recreate them',
+        )
+
+    def handle(self, *args, **options):
+        irods_backend = plugin_api.get_backend_api('omics_irods')
+        default_admin = User.objects.get(
```
mikkonie, line 130: For consistency, I'd like us to have info level log lines for operation start and end, with e.g. the count of objects created. See e.g. `syncnames` for an example of this.

Before you say it: yes, there is also a case to be made for unix-style "golden rule of silence" :) We could consider doing that in the future unless a verbose option was given. For now, I'd prefer providing the admin running the command some concrete feedback on the operation by default.

samplesheets/management/commands/fixirodstickets.py
```diff
@@ -0,0 +1,114 @@
...
+        check = options.get('check')
+        for ticket, path in self._find_irods_tickets(irods_backend):
+            ticket_string = ticket[TicketQuery.Ticket.string]
+            try:
+                IrodsAccessTicket.objects.get(ticket=ticket_string)
+                logger.info(f'Found existing ticket object for {ticket_string}')
+            except IrodsAccessTicket.DoesNotExist:
+                ticket_study = irods_backend.get_uuid_from_path(path, 'study')
+                if ticket_study:
+                    ticket_study = Study.objects.get(sodar_uuid=ticket_study)
```
mikkonie: In case of an orphaned study collection, this may raise a `DoesNotExist` exception. This should be handled, logged and tested.

samplesheets/management/commands/fixirodstickets.py
```diff
@@ -0,0 +1,114 @@
...
+                else:
+                    logger.warning(
+                        f'Ticket {ticket_string} does not have '
+                        'an associated study'
+                    )
+                    continue
+                ticket_assay = irods_backend.get_uuid_from_path(path, 'assay')
+                if ticket_assay:
+                    ticket_assay = Assay.objects.get(sodar_uuid=ticket_assay)
```
mikkonie: Same thing here as above with the potentially missing `Study` object.

samplesheets/management/commands/fixirodstickets.py
```diff
@@ -0,0 +1,114 @@
...
+                    study=ticket_study,
+                    assay=ticket_assay,
+                    ticket=ticket_string,
+                    path=path,
+                    user=default_admin,
+                    date_created=ticket[TicketQuery.Ticket.create_time],
+                    date_expires=ticket_expires,
+                )
+                logger.info(
+                    f'Need to recreate ticket object for {ticket_string} (\n'
```
mikkonie: If check is not enabled, the log should state "Created ticket object for.." after we have performed the creation.

The additional information regarding exact model fields isn't so critical here IMHO, and I'd prefer to keep single log entries as single lines for ease of parsing. Instead, I would just log the project and UUID of created object.

Something like this:

`Created database object for ticket {ticket_string} in {project.get_log_title()} (UUID={obj.sodar_uuid})`.

That would be consistent with the general style of logging in commands.

samplesheets/tests/test_commands_taskflow.py
```diff
@@ -202,3 +218,130 @@ def test_sync_public_guest_viewer_revoke(self):
...
+class TestFixIrodsTickets(
+    IrodsAccessTicketMixin, SampleSheetIOMixin, TaskflowViewTestBase
+):
+    """Tests for the fixirodstickets command"""
+
+    def setUp(self):
+        super().setUp()
+        self.admin = self.make_user(settings.PROJECTROLES_DEFAULT_ADMIN)
+        # Create project locally
+        self.project = self.make_project(
```
mikkonie: I recommend using `make_project_taskflow()` here, as it ensures we have set up the project in a fashion we would have if we would be introducing iRODS tickets to it in the first place.

samplesheets/tests/test_commands_taskflow.py
```diff
@@ -202,3 +218,130 @@ def test_sync_public_guest_viewer_revoke(self):
...
+            'NewProject', PROJECT_TYPE_PROJECT, self.category
+        )
+        self.make_assignment(self.project, self.user, self.role_owner)
+        # Import investigation and create iRODS objects
+        self.investigation = self.import_isa_from_file(SHEET_PATH, self.project)
+        self.study = self.investigation.studies.first()
+        self.assay = self.study.assays.first()
+        self.coll_path = self.irods_backend.get_path(self.assay)
+        self.irods.collections.create(self.coll_path)
```
mikkonie: Here I recommend using `SampleSheetTaskflowMixin.make_irods_colls()`, again to ensure we have the project iRODS data set up exactly the way we would in a real life use case.

mikkonie (changes requested):
Looking good, well tested and documented too. I have some suggestions for improvement, mostly minor things.

samplesheets/management/commands/fixirodstickets.py
fmarotta: OK, I've removed the type hint.

samplesheets/management/commands/fixirodstickets.py
fmarotta: Done

samplesheets/management/commands/fixirodstickets.py
fmarotta: Log lines added.

samplesheets/management/commands/fixirodstickets.py
fmarotta: Good point, done

samplesheets/management/commands/fixirodstickets.py
fmarotta: Also done

samplesheets/management/commands/fixirodstickets.py
fmarotta: Sounds good. Just to clarify, the original issue said that each found/recreated object should be logged: should the found tickets be logged only if check==False? This is what I've assumed, but let me know.

samplesheets/tests/test_commands_taskflow.py
fmarotta: Good point, I'll keep this in mind for next time.

samplesheets/tests/test_commands_taskflow.py
fmarotta: Done

samplesheets/management/commands/fixirodstickets.py
mikkonie: Yes, everything should be logged. The text should simply be different based on whether we recreated the ticket or simply found an orphaned one.

samplesheets/management/commands/fixirodstickets.py
```diff
@@ -0,0 +1,143 @@
...
+                yield (
+                    ticket,
+                    iRODSPath(
+                        ticket[TicketQuery.DataObject.coll],
+                        ticket[TicketQuery.DataObject.name],
+                    ),
+                )
+
+    @classmethod
+    def _extract_ticket_properties(
```
mikkonie: We should use consistent wording in functions, so `_get*()` should be OK here. Also, since we are explicitly constructing Django model fields, I propose naming this `_get_ticket_fields()` to remove any ambiguity.

samplesheets/management/commands/fixirodstickets.py
```diff
@@ -0,0 +1,143 @@
...
+        cls, irods_backend, ticket, path
```
mikkonie: It would be nice to have type hints and and a docstring here. The type hint for `irods_backend` can just be `Any` since we don't want to add hardcoded imports here.

samplesheets/management/commands/fixirodstickets.py
```diff
@@ -0,0 +1,143 @@
...
+                if not check:
+                    obj.save()
+                    logger.info(
+                        f'Created database object for ticket {ticket_string} '
+                        f'in {obj.study.investigation.project.get_log_title()} '
+                        f'(UUID={obj.sodar_uuid})'
+                    )
+        logger.info(
+            ('Found' if check else 'Recreated')
+            + f' {orphaned_tickets_count} orphaned tickets objects.'
```
mikkonie: This should probably say "ticket objects".

samplesheets/tests/test_commands_taskflow.py
```diff
@@ -202,3 +217,171 @@ def test_sync_public_guest_viewer_revoke(self):
...
+    SampleSheetTaskflowMixin,
+    SampleSheetIOMixin,
+    TaskflowViewTestBase,
+):
+    """Tests for the fixirodstickets command"""
+
+    def setUp(self):
+        super().setUp()
+        self.admin = self.make_user(settings.PROJECTROLES_DEFAULT_ADMIN)
+        # Create project locally
```
mikkonie: We are not creating the project locally anymore as we use `make_project_taskflow()`, so this can say e.g. "Create project with taskflow".

samplesheets/tests/test_commands_taskflow.py
```diff
@@ -202,3 +217,171 @@ def test_sync_public_guest_viewer_revoke(self):
...
+            IrodsAccessTicket.objects.filter(ticket=DATA_TICKET_STR).count(),
+            1,
+        )
+        self.study.delete()
+        self.assertEqual(
+            IrodsAccessTicket.objects.filter(ticket=DATA_TICKET_STR).count(),
+            0,
+        )
+        call_command('fixirodstickets')
+        # Nothing should happen to the database, but a warning will be shown
```
mikkonie, line 387: In cases where we can't determine the outcome directly from the object output, I recommend asserting the relevant log lines are displayed using `self.assertLogs()`. You should find examples of this elsewhere in the repo.

I also recommend setting those strings as constants in the command and importing them here to avoid repetition.

mikkonie (changes requested):
Looking good. I have a couple of very minor requests and a suggestion for improving the tests.

samplesheets/management/commands/fixirodstickets.py
fmarotta: Done

samplesheets/management/commands/fixirodstickets.py
fmarotta: :+1:

samplesheets/management/commands/fixirodstickets.py
fmarotta: I fixed the typo and added a condition to remove the "s" when the ticket count is 1.

samplesheets/tests/test_commands_taskflow.py
fmarotta: Done

samplesheets/tests/test_commands_taskflow.py
fmarotta: Good point, I wasn't aware of that function

samplesheets/management/commands/fixirodstickets.py
```diff
@@ -0,0 +1,168 @@
...
+                    logger.info(
+                        'Found existing object for ticket {ticket_string}'
+                    )
+            except IrodsAccessTicket.DoesNotExist:
+                ticket_fields = self._get_ticket_fields(
+                    irods_backend, ticket, path
+                )
+                if not ticket_fields:
+                    continue
+                obj = IrodsAccessTicket(
```
mikkonie: If `check` mode is enabled, we should not create this object. The check mode should not modify anything.

samplesheets/management/commands/fixirodstickets.py
```diff
@@ -0,0 +1,168 @@
...
+                    ticket=ticket_string,
+                    path=path,
+                    user=default_admin,
+                    **ticket_fields,
+                )
+                orphaned_ticket_count += 1
+                if not check:
```
mikkonie: We should log found orphaned tickets regardless of `check` mode. Of course, the message should then be modified to say something else than "created database object", as the check mode should not create anything.

In the case of `check` enabled, the message should be something like "Found orphaned ticket..". Of course, the ticket object UUID will also not be present.

Sorry if I was unclear about this before.

mikkonie (changes requested):
Good work adressing my comments. I found two issues related to check mode, probably due to my unclear specs. If you can still look into these, I'm sure we can merge in the next round :)

samplesheets/management/commands/fixirodstickets.py
fmarotta: Done. The object should not be created in the db until the `.save()` method is called, but I agree that instantiating it only in the "not check" branch is clearer.

samplesheets/management/commands/fixirodstickets.py
fmarotta: Ah, yes, I had misinterpreted the spec. Orphaned tickets are now logged in check mode. I have another question (sorry for possibly prolonging this to another round, but I want to make sure to get it right :)). What about existing ticket objects? Should they be logged or not?

mikkonie: I think it's enough to just log the orphaned ones.

mikkonie (approved):
Good work! To be merged :)

Issue #2394, opened 2026-02-13
Disable landing zone create form assay dropdown if only one assay is available


mikkonie:
Minor issue, but we should make things straightforward to the user. If there is only one selectable assay for the landing zone, the widget should be disabled. The existing assay should be preselected, but that is already the case, so nothing further should be needed there.

Testing this in view tests should be enough, to ensure the form widget is disabled.

## Tasks

- [ ] Disable `assay` field dropdown widget if only one assay is available
- [ ] Ensure this in view tests

mikkonie:
Done by @fmarotta, merged.

PR #2492, opened 2026-08-03
Disable landing zone create form assay dropdown if only one assay is available

fmarotta:
This PR addresses #2394.


mikkonie (approved):
Looks good, approved.

I think the test could be simplified by just adding a new `Assay` object to the existing project and creating an iRODS collection for it, but admittedly this is definitely a more realistic setup :)

Issue #2144, opened 2025-04-29
Maligned tooz lock ID generation in ProjectLockAPI


mikkonie:
I thought I already had opened a ticket for this, but guess not. While doing #2048 I noticed it again and this should be handled now and there.

Can you spot the very much invalid ID?

```
b'celery-task-meta-c1d4bc5d-2df2-4b51-9f23-f8e921df0f47'
b'celery-task-meta-054c0880-8087-4db0-a4d6-64c71f3dd79f'
b'_kombu.binding.celery'
b'_kombu.binding.celery.pidbox'
b'celery-task-meta-5c2aa8e7-3942-4b45-90ee-73d098779604'
b"b'_tooz'_bf321abe-36c1-4ce4-be15-ce61c98fda66_lock"
b'_tooz_beats:sodar_a22412ee-53fc-486c-a7fc-6b9adc302104'
b'celery-task-meta-1910de66-f534-486c-b106-1cb0341341fa'
b'_tooz_beats:sodar_3f6a36b1-d288-473b-a077-877f125512ea'
b'_kombu.binding.celeryev'
b'celery-task-meta-ab236922-c8a0-41db-88f5-f14a7cc37817'
b'celery-task-meta-8df2c422-e8d2-4c75-a3e3-cd57985be914'
b'celery-task-meta-cef45bc2-6b55-44bc-aaec-f11f422f494c'
b'celery-task-meta-f9d875ad-ae5b-4229-8018-b384ddd93ebd'
```

Of course the ultimate solution might be to replace tooz entirely as it's a bit overkill for our current use case, see #1844. But that's something for a later milestone, let's get this fixed now.

mikkonie:
This actually seems like a bug in tooz or its redis plugin. We only provide an ID as a string to `coordinator.get_lock()` which then constructs the maligned "double bytestring" ID.

I can't really see how to control this. I tried providing the ID directly as bytestring, but that doesn't help.

I upgraded tooz to the latest release just in case (SODAR was running quite an old version), but that did not affect the outcome.

I'll probably need to raise an issue regarding this in the tooz repository, if no one else has yet done this.

This is not really a huge issue as the Redis database IDs are not accessed directly (apart for the hack I did for #2048) but it sure doesn't help debugging..

mikkonie:
Fixed by @fmarotta by upgrading to Tooz v9 in #2485.

Issue #2485, opened 2026-07-22
Upgrade to tooz v9


fmarotta:
After upgrading the tooz package to the latest version, some `taskflowbackend` tests fail with the following error:

```
taskflowbackend.api.TaskflowAPI.FlowSubmitException: Unable to acquire project lock: a bytes-like object is required, not 'str'
```

the [release notes](https://docs.openstack.org/releasenotes/tooz/unreleased.html) may offer some clues..

mikkonie:
I can have a look since I've worked with tooz quite a bit. Should probably be easy to fix.

At some point we might consider getting rid of tooz entirely, see #1844. However, that's for the future.

mikkonie:
Done by @fmarotta, merged.

PR #2503, opened 2026-08-10
Upgrade tooz to v9.0

fmarotta:
This PR addresses #2485. After upgrading, the lock name is returned as a `bytes` object, which made `_log_status()` fail. It seems that the internal encoding of the lock id has also changed: I checked by printing the keys in redis and I didn't see `"b'_tooz'_{project.sodar_uuid}_lock"` but rather `'_tooz_{project.sodar_uuid}_lock'`.


mikkonie (approved):
Good work, will merge. It's nice to see that the Openstack people finally fixed that issue with the lock names. This also closes #2144.

Issue #2510, opened 2026-08-14
Add plugin search() tests


mikkonie:
These seem to be totally missing for some reason. Especially with the renewed search implementation in SODAR Core v1.4, it would be very useful to have these in place.

## Spec

- Add tests for `ProjectAppPlugin.search()` for landingzones and samplesheets
- Similar to plugin tests in SODAR Core
- Test for all search types (of course)
- Samplesheets tests involving iRODS files should go under `test_plugins_taskflow` with taskflow/iRODS setup
    * Everything else should work under simple `test_plugins` files with non-taskflow init
- Existing tests should provide plenty of examples on how to set up sample sheets for tests, but feel free to ask for hints :)

## Tasks

- [ ] Add landingzones tests
- [ ] Add samplesheets tests
- [ ] Add samplesheets tests for irods file search with taskflow setup

mikkonie:
Done by @fmarotta, merged.

PR #2528, opened 2026-08-28
Add plugin search() tests

fmarotta:
This PR addresses #2510

### DIFF
--- a/samplesheets/tests/test_plugins.py
+++ b/samplesheets/tests/test_plugins.py
@@ -65,7 +65,7 @@ PROJECT_TYPE_CATEGORY = SODAR_CONSTANTS['PROJECT_TYPE_CATEGORY']
 APP_NAME = 'samplesheets'
 SHEET_PATH = SHEET_DIR + 'i_minimal2.zip'
 MATERIAL_NAME = '0815-N1-DNA1'
-SOURCE_NAME = '0815-N1'
+SAMPLE_NAME = '0815-N1'
 ASSAY_PLUGIN_NAME = 'samplesheets.assayapps.dna_sequencing'
 SHEET_COL_ID = 'sheets'
 FILE_COL_ID = 'files'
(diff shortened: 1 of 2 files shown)


samplesheets/tests/test_plugins.py
```diff
@@ -56,6 +65,7 @@
 APP_NAME = 'samplesheets'
 SHEET_PATH = SHEET_DIR + 'i_minimal2.zip'
 MATERIAL_NAME = '0815-N1-DNA1'
+SOURCE_NAME = '0815-N1'
```
mikkonie: While this obviously doesn't affect the execution of tests, in our general notation `*-N1` commonly denotes a sample taken from a source. Hence, in this case the name of a source would usually be something like `0815`.

samplesheets/tests/test_plugins.py
```diff
@@ -422,3 +432,198 @@ def test_update_invalid_path(self):
...
+        self.assay2 = self.study2.assays.first()
+        # Create the same sample in the project2
+        GenericMaterial.objects.create(
+            name='0816-N1',
+            unique_name='0816-N1-1-1',
+            item_type=ITEM_TYPE_SAMPLE,
+            study=self.study2,
+        )
+
+    def _test_material_row_contents(self, row, name, type, study, assay=None):
```
mikkonie: Just a couple of minor style things here:

- In test classes, I prefer to declare internal helper functions before `setUp()`.
- Since this is not a test case in itself but an assertion helper, let's preface the function name with `_assert*` instead of `_test*`.

samplesheets/tests/test_plugins_taskflow.py
```diff
@@ -934,3 +941,212 @@ def test_get_category_stats_file(self):
...
+        self.assay2 = self.study2.assays.first()
+        self.make_irods_colls(self.investigation2)
+        self.misc_path2 = iRODSPath(
+            self.irods_backend.get_path(self.assay2), MISC_FILES_COLL
+        )
+        self.irods.collections.create(self.misc_path2)
+        misc_coll2 = self.irods.collections.get(self.misc_path2)
+        self.irods_misc_obj2 = self.make_irods_object(misc_coll2, TEST_OBJ_NAME)
+
+    def _test_file_row_contents(self, row, name, project, user, assay):
```
mikkonie: Same function ordering/naming comments here as for the helper in the other test class.

mikkonie (changes requested):
Good work! I can't immediately think of any realistic cases not tested here.

I had a couple very small requests for tweaks, once those are done I'm happy to merge.

samplesheets/tests/test_plugins.py
fmarotta: I see! I have renamed the variable to `SAMPLE_NAME`.

samplesheets/tests/test_plugins.py
fmarotta: Done

samplesheets/tests/test_plugins_taskflow.py
fmarotta: Done

mikkonie (approved):
Looking great, will merge!

